PR #7, opened 2020-05-13
Circuit method for determinstic square roots

dlubarov:
To enforce determinism, we require the square root to have a parity bit of zero.

The generator (prover) has a few steps:
- Call `Field::square_root` to compute any square root.
- If the parity bit is 1, negate it.
- Call `split_binary_and_base_4` on the square root to populate the bit and dibit wires.

The verifier involves these steps:
- Square the purported square root and make sure the result matches.
- Check that the parity bit is 0.
- Range check each limb (`Base4SumGate` range checks the base 4 limbs).
- Verify that a weighted sum of the binary and base 4 limbs matches the original input.

It a little tricky to do this in a way that is agnostic to field size. In particular, we want the number of dibits to be a multiple of 7, since that's how many dibits Base4SumGate can handle. We would need to rework things a bit if we wanted to sum up an arbitrary number of dibits in an efficient and secure way. So for now, I'm just asserting that the field size is ~254 bits.

### DIFF
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 use std::time::Instant;
 
 use crate::{AffinePoint, Curve, Field, generate_rescue_constants, HaloEndomorphismCurve};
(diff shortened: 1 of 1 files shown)


dlubarov:
Sean also suggested a more clever method in zcash/zcash#4421, but this approach seems fairly efficient in our circuit model (18 `Base4Sum` gates and a couple other gates).

PR #11, opened 2020-05-18
adding method to check whether a vector of limbs is a valid element o…

bpfarmer:
…f a field

### DIFF
--- a/src/tweedledee_base.rs
+++ b/src/tweedledee_base.rs
@@ -1,12 +1,15 @@
+use rand::Rng;
 use std::cmp::Ordering::Less;
 use std::convert::TryInto;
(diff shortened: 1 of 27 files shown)


src/tweedledee_base.rs
```diff
@@ -309,5 +313,19 @@ mod tests {
         }
     }
 
+    #[test]
+    fn valid_canonical_vec() {
+        let small = TweedledeeBase::ONE.to_canonical_u64_vec();
+        assert!(TweedledeeBase::is_valid_canonical_u64(small));
+
+
+        let mut big = Vec::new();
```
dlubarov, line 322: nit: can be `let big = TweedledeeBase::ORDER_X2.to_vec()`

PR #12, opened 2020-05-19
Prover groundwork

dlubarov:
- Move `ProofTarget` into a separate file, and add `Proof` which is similar, but with concrete curve/field data instead of `Target`s (the latter being for recursive verification).
- Parameterize `CircuitBuilder` with the curve type, since its `build` method will do preprocessing involving the curve, like committing to constant polynomials.


dlubarov:
@bpfarmer @wborgeaud just FYI

PR #13, opened 2020-05-20
Faster hash to curve

wborgeaud:
Implemented a hash-to-curve method using the same mechanics as the previous one, but using Blake3 instead of Rescue for performance reasons.


src/hash_to_curve.rs
```diff
@@ -58,4 +125,37 @@ mod tests {
             assert_eq!(x, hash_u32_to_curve::<Tweedledum>(i, 128));
         }
     }
+
+    #[test]
+    fn test_hash_rescue_bench() {
```
dlubarov, line 130: Are these still useful since you added the benchmark?

PR #16, opened 2020-05-22
More prover work

dlubarov:
This is sort of a sketch of the prover code; there are a lot of TODOs to be filled in later.

I'd like to break up this huge generate_proof method, but need to think some more about how to do it cleanly.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -76,7 +76,7 @@ pub trait Field:
         limbs
     }
 
-    fn from_canonical_u8_vec(u8_limbs: Vec<u8>) -> Result<Self> {
+    fn from_canonical_u8_vec(u8_limbs: Vec<u8>) -> Option<Self> {
         let mut u64_chunks = Vec::new();
         for u8_chunk in u8_limbs.chunks(8) {
             u64_chunks.push(
@@ -90,10 +90,11 @@ pub trait Field:
                     | (u8_chunk[0] as u64),
             );
         }
+
         if Self::is_valid_canonical_u64(&u64_chunks) {
-            Ok(Self::from_canonical_u64_vec(u64_chunks))
+            Some(Self::from_canonical_u64_vec(u64_chunks))
         } else {
-            Err(Error::msg("lol"))
+            None
         }
     }
 
@@ -182,6 +183,10 @@ pub trait Field:
     #[inline(always)]
     fn quadruple(&self) -> Self {
         *self * Self::FOUR
+    }
+
+    fn inner_product(a: &[Self], b: &[Self]) -> Self {
+        todo!()
     }
 
     fn batch_multiplicative_inverse_opt<F: Field>(x: &[F]) -> Vec<Option<F>> {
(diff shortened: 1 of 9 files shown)


src/field.rs
```diff
@@ -76,7 +76,7 @@ pub trait Field:
         limbs
     }
 
-    fn from_canonical_u8_vec(u8_limbs: Vec<u8>) -> Result<Self> {
+    fn from_canonical_u8_vec(u8_limbs: Vec<u8>) -> Option<Self> {
```
dlubarov, line 79: (Doesn't really matter, but Option seems slightly more natural to me when there's just one failure case and the caller is expected to understand it. Let me know if you prefer with Result though.)

src/plonk.rs
```diff
@@ -106,6 +120,184 @@ impl<C: Curve> Circuit<C> {
...
+        let all_opening_sets: Vec<OpeningSet<C::ScalarField>> = [
+            o_public_inputs,
+            vec![o_local, o_right, o_below],
+        ].concat();
+        let all_opened_values_sf: Vec<C::ScalarField> = all_opening_sets.into_iter()
+            .map(|os| os.to_vec()).collect::<Vec<_>>()
+            .concat();
+        let all_opened_values_bf: Vec<_> = all_opened_values_sf.into_iter()
+            .map(|f| C::try_convert_s2b(f)
+                .expect("For now, we assume that all opened values fit in both fields"))
```
dlubarov, line 210: This is a pretty big limitation -- the prover may fail if we try to use -1 as a public input (depending on which field is larger), for example. I'll try to address this soon, but assuming all openings fit in both fields makes things a lot easier for now.

src/field.rs
wborgeaud: As you prefer, I tend to like `Result`s when there is a failure involve so that I can add a meaningful error message for future implementers.
But in this specific case,  an `Option` is perfectly fine since the code is very simple, and is more coherent with the rest of the code.

wborgeaud:
Looks good to me. I agree that it would be good to break the function up later.
Also, maybe a `Transcript` struct could be nice to abstract away Fiat-Shamir?

src/field.rs
dlubarov: Yeah that's a fair point. I'm really on the fence in this case, would be happy with either one.

wborgeaud: Also I think for complex functions (e.g. `generate_proof`) it might be cleaner to return a `Result` and use the `?` operator whenever an operation fails. If we use `Option`s, we'd have to `unwrap` all possibly failing operations or write a custom message in `except`, which is imo a bit less clean.

dlubarov:
I like the idea of a Fiat-Shamir abstraction -- I took a stab at it and the generate_proof looks quite a bit cleaner now :)

src/field.rs
dlubarov: Ah yeah that could be nice. Switching it back.

PR #18, opened 2020-05-23
Generate k_i, the subgroup shifts used in Plonk's permutation argument

dlubarov:


src/plonk_recursion.rs
```diff
@@ -305,14 +305,9 @@ fn verify_assumptions<C: Curve, InnerC: HaloEndomorphismCurve<BaseField=C::Scala
     // Compute Z(zeta) f'(zeta) - Z(g * zeta) g'(zeta), which should vanish on H.
     let mut f_prime = one;
     let mut g_prime = one;
-    let quadratic_nonresidues = C::ScalarField::generate_quadratic_nonresidues(NUM_ROUTED_WIRES - 1);
     for i in 0..NUM_ROUTED_WIRES {
-        let s_id = if i == 0 {
-            zeta
-        } else {
-            let k_i = builder.constant_wire(quadratic_nonresidues[i - 1]);
```
dlubarov, line 313: (I wrote this old code just based on the following, without thinking much about it --

> This can be done by defining `S_ID_1(X) = X`, `S_ID_2(X) = k_1 X`, `S_ID_3(X) = k_2 X`, where `k_1`, `k_2` are quadratic non-residues ∈ `F`.

-- but in hindsight, to be safe I think it should have checked that each `k_i` after the first two was not in a previous coset. Anyway, I think using random shifts is safe and simpler.)

src/plonk.rs
```diff
@@ -152,16 +156,25 @@ impl<C: Curve> Circuit<C> {
 
         // Generate a random beta and gamma from the transcript.
         challenger.observe_affine_points(&c_wires);
-        let (beta, gamma) = challenger.get_2_challenges();
+        let (beta_bf, gamma_bf) = challenger.get_2_challenges();
+        let beta_sf = beta_bf.try_convert::<C::ScalarField>()?;
+        let gamma_sf = gamma_bf.try_convert::<C::ScalarField>()?;
+
+        let subgroup = C::ScalarField::cyclic_subgroup_known_order(self.subgroup_generator, self.degree());
```
dlubarov, line 163: Note that this list of subgroup elements starts with `1`, so the code below will start with `Z(1) = 1`. The Plonk paper starts with `Z(g) = 1`, although I think other parts of the paper assume `Z(1) = 1`. (It's hard to tell since `L_1(x)` could refer to either the function that selects `1` or `g^1`; both come up in the paper.)

Of course either works, we just need to make sure we're consistent, and keep in mind that the paper won't always match.

wborgeaud:
LGTM. I like the idea of choosing a random shift.

PR #21, opened 2020-05-28
Division by Z_h

wborgeaud:
Added a method to divide a polynomial by `Z_H = X^n - 1` using coset FFT.


src/plonk.rs
```diff
@@ -202,8 +202,7 @@ impl<C: Curve> Circuit<C> {
             &wire_values_8n, alpha_sf, beta_sf, gamma_sf, &plonk_z_coeffs);
 
         // Compute the quotient polynomial, t(x) = vanishing(x) / Z_H(x).
-        let (plonk_t_coeffs, r): (Vec<C::ScalarField>, Vec<C::ScalarField>) = polynomial_division(&vanishing_coeffs, &plonk_z_coeffs);
-        assert_eq!(r, Vec::new());
+        let plonk_t_coeffs: Vec<C::ScalarField> = divide_by_z_h(&vanishing_coeffs, self.degree());
```
wborgeaud, line 205: Is `self.degree()` correct here, or is `n` a different number?

dlubarov: Oh yeah that's right -- by degree I meant degree of H. Do you think it would be more clear if we rename to something like `gate_count()` or maybe just `n()`?

wborgeaud: Oh no it's ok, `degree` is fine, I just wanted to be sure.

PR #24, opened 2020-06-01
Fixes related to gate_constants

dlubarov:
gate_constants is index by gate index and then wire index, but the code that IFFTs it expected the opposite.


dlubarov:
My goal there was to get timing data for recursive proof generation, but then I realized it would take a fair amount of code to generate the proof data that the recursive circuit expects as inputs. I'll send a followup PR to finish that soon!

PR #29, opened 2020-06-05
Fixes to get main method working

dlubarov:


src/circuit_builder.rs
```diff
@@ -1,11 +1,6 @@
 use std::collections::{BTreeMap, HashMap};
 
-use crate::{
-    AffinePoint, AffinePointTarget, blake_hash_usize_to_curve, Circuit, Curve, CurveMsmEndoResult,
-    CurveMulEndoResult, CurveMulOp, fft_precompute, Field, generate_rescue_constants, HaloCurve,
-    msm_precompute, NUM_CONSTANTS, NUM_WIRES, PartialWitness, PublicInput, Target, TargetPartitions,
-    VirtualTarget, Wire, WitnessGenerator,
-};
+use crate::{AffinePoint, AffinePointTarget, blake_hash_usize_to_curve, Circuit, Curve, CurveMsmEndoResult, CurveMulEndoResult, CurveMulOp, fft_precompute, Field, generate_rescue_constants, HaloCurve, msm_precompute, NUM_CONSTANTS, NUM_WIRES, PartialWitness, PublicInput, Target, TargetPartitions, VirtualTarget, Wire, WitnessGenerator, blake_hash_base_field_to_curve};
```
dlubarov, line 3: @wborgeaud are you using IntelliJ or something else? I don't have any preference on formatting things like this, just IntelliJ seems to collapse the lines when I import something.

wborgeaud: @dlubarov  I'm using VS Code with rust-analyzer, which uses rustfmt for formating I think. It's probably something that I can configure though, I'll look into it.

dlubarov: Oh ok, I also checked in IntelliJ's Rust formatting settings but didn't see options for that. I've been meaning to try rust-analyzer though so maybe we can standardize that. (Or we can just let it change back and forth -- no big deal, I just wanted to explain.)

wborgeaud: I found how to set custom configs in rustfmt in a file .rustfmt.toml that I just pushed on master. I think it's possible to use rustfmt with IntelliJ, so maybe we can standardize using rustfmt and customize .rustfmt.toml as we go.

dlubarov: Sounds good to me 👍

PR #34, opened 2020-06-11
Add tests for low-degreeness of each constraint

dlubarov:
I `#[ignore]`d the tests since they're expensive, but they can be run like this:

    cargo test low_degree -- --ignored

### DIFF
--- a/src/plonk_gates.rs
+++ b/src/plonk_gates.rs
@@ -1408,10 +1408,10 @@ macro_rules! test_gate_low_degree {
 
             // Make sure each extended polynomial is still degree n.
             for values_8n in &constant_values_8n {
-                assert_eq!($crate::plonk_util::polynomial_degree(values_8n, &fft_precomputation_8n), n);
+                assert_eq!($crate::plonk_util::polynomial_degree(values_8n, &fft_precomputation_8n), n - 1);
             }
             for values_8n in &wire_values_8n {
-                assert_eq!($crate::plonk_util::polynomial_degree(values_8n, &fft_precomputation_8n), n);
+                assert_eq!($crate::plonk_util::polynomial_degree(values_8n, &fft_precomputation_8n), n - 1);
             }
 
             let constant_values_8n_t = $crate::util::transpose(&constant_values_8n);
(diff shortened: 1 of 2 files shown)


src/plonk_gates.rs
```diff
@@ -1373,3 +1373,89 @@ impl<C: HaloCurve> WitnessGenerator<C::ScalarField> for ArithmeticGate<C> {
...
+            let wire_coeffs_n = $crate::plonk_util::values_to_coeffs(&wire_values_n, &fft_precomputation_n);
+            let wire_values_8n = $crate::plonk_util::coeffs_to_values_padded(&wire_coeffs_n, &fft_precomputation_8n);
+
+            // Make sure each extended polynomial is still degree n.
+            for values_8n in &constant_values_8n {
+                assert_eq!($crate::plonk_util::polynomial_degree(values_8n, &fft_precomputation_8n), n - 1);
+            }
+            for values_8n in &wire_values_8n {
+                assert_eq!($crate::plonk_util::polynomial_degree(values_8n, &fft_precomputation_8n), n - 1);
+            }
```
wborgeaud, line 1415: Shouldn't this be `assert!($crate::plonk_util::polynomial_degree(values_8n, &fft_precomputation_8n) <= n - 1);` instead? These polynomial should be of degree `n-1` with overwhelming probability though, so it shoudn't matter.

src/plonk_util.rs
```diff
@@ -244,6 +244,14 @@ pub fn sigma_polynomials<F: Field>(
...
 }
 
+pub(crate) fn polynomial_degree<F: Field>(
+    points: &[F],
+    fft_precomputation: &FftPrecomputation<F>,
+) -> usize {
+    let coeffs = ifft_with_precomputation_power_of_2(&points, fft_precomputation);
+    coeffs.iter().rev().skip_while(|c| c.is_zero()).count() - 1
+}
+
```
wborgeaud, line 254: I'm not sure how much it matters in practice, but this can give incorrect results for the degree. For example the polynomial `X^16n` will have degree 0 according to this function. I think we can add a check that the evaluation at a random point outside of `H` is consistent with the coefficients returned by the iFFT to be sure that the degree is less than `8n`.

src/plonk_gates.rs
dlubarov: Good point -- fixed.

src/plonk_util.rs
dlubarov: Whoops -- so my intention was to evaluate constraints over a larger domain, check the degree of the interpolant, and assert that it's within the max degree. But I was using the 8n subgroup as my "larger domain", and since the current max degree is also 8n, I wasn't catching anything.

I changed this to use a 16n subgroup, and a few of these tests (like Base4Sum's) are now failing as expected. Hopefully it makes sense now?

PR #35, opened 2020-06-12
Fix CurveDblGate

dlubarov:


dlubarov:
I ran `cargo test low_degree_CurveDblGate -- --ignored` to confirm that it passes now.

PR #38, opened 2020-06-14
Verifier

wborgeaud:
Contains the bulk of the verifier work and some bug fixes.

### DIFF
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -6,11 +6,19 @@ use anyhow::Result;
 
 use crate::partition::{get_subgroup_shift, TargetPartitions};
 use crate::plonk_challenger::Challenger;
-use crate::plonk_util::{coeffs_to_commitments, coeffs_to_values_padded, eval_coeffs, eval_l_1, eval_poly, eval_zero_poly, halo_n, pad_to_8n, pedersen_hash, permutation_polynomial, powers, reduce_with_powers, values_to_coeffs};
+use crate::plonk_util::{
+    coeffs_to_commitments, coeffs_to_values_padded, eval_coeffs, eval_l_1, eval_poly,
+    eval_zero_poly, halo_n, pad_to_8n, pedersen_hash, permutation_polynomial, powers,
+    reduce_with_powers, values_to_coeffs,
+};
 use crate::target::Target;
 use crate::util::{ceil_div_usize, log2_strict};
 use crate::witness::{PartialWitness, Witness, WitnessGenerator};
-use crate::{divide_by_z_h, evaluate_all_constraints, fft_with_precomputation_power_of_2, ifft_with_precomputation_power_of_2, msm_parallel, AffinePoint, CircuitBuilder, Curve, FftPrecomputation, Field, HaloCurve, MsmPrecomputation, OpeningSet, ProjectivePoint, Proof};
+use crate::{
+    divide_by_z_h, evaluate_all_constraints, fft_with_precomputation_power_of_2,
+    ifft_with_precomputation_power_of_2, msm_parallel, AffinePoint, CircuitBuilder, Curve,
+    FftPrecomputation, Field, HaloCurve, MsmPrecomputation, OpeningSet, ProjectivePoint, Proof,
+};
 
 pub(crate) const NUM_WIRES: usize = 9;
 pub(crate) const NUM_ROUTED_WIRES: usize = 6;
@@ -213,9 +221,10 @@ impl<C: HaloCurve> Circuit<C> {
 
         // Generate random v, u, and x from the transcript.
         challenger.observe_elements(&all_opened_values_bf);
-        let (v_bf, u_bf, _x) = challenger.get_3_challenges();
+        let (v_bf, u_bf) = challenger.get_2_challenges();
         let v_sf = v_bf.try_convert::<C::ScalarField>()?;
         let u_sf = u_bf.try_convert::<C::ScalarField>()?;
+        let u_curve = challenger.get_affine_point::<C>();
 
         // Make a list of all polynomials' commitments and coefficients, to be reduced later.
         // This must match the order of OpeningSet::to_vec.
@@ -279,6 +288,8 @@ impl<C: HaloCurve> Circuit<C> {
         let mut halo_g = AffinePoint::batch_to_projective(&self.pedersen_g);
         let mut halo_l = Vec::new();
         let mut halo_r = Vec::new();
+        // TODO: Change this to the real randomness used in the polynomial commitments.
+        let mut randomness = C::ScalarField::ZERO;
         for j in (1..=self.degree_pow()).rev() {
             let n = 1 << j;
             let middle = n / 2;
@@ -318,6 +329,10 @@ impl<C: HaloCurve> Circuit<C> {
             let l_challenge_sf = l_challenge_bf.try_convert::<C::ScalarField>()?;
             let r_challenge_sf = r_challenge_bf.try_convert::<C::ScalarField>()?;
 
+            randomness = randomness
+                + l_challenge_sf.square() * l_j_blinding_factor
+                + r_challenge_sf.square() * r_j_blinding_factor;
+
             halo_a = C::ScalarField::add_slices(
                 &l_challenge_sf.scale_slice(a_lo),
                 &r_challenge_sf.scale_slice(a_hi),
@@ -335,6 +350,10 @@ impl<C: HaloCurve> Circuit<C> {
         debug_assert_eq!(halo_g.len(), 1);
         let halo_g = halo_g[0].to_affine();
 
+        debug_assert_eq!(halo_a.len(), 1);
+        debug_assert_eq!(halo_b.len(), 1);
+        let schnorr_proof = self.schnorr_protocol(halo_a[0], halo_b[0], halo_g, randomness, u_curve, &mut challenger);
+
         Ok(Proof {
             c_wires,
             c_plonk_z,
@@ -346,6 +365,7 @@ impl<C: HaloCurve> Circuit<C> {
             halo_l: ProjectivePoint::batch_to_affine(&halo_l),
             halo_r: ProjectivePoint::batch_to_affine(&halo_r),
             halo_g,
+            schnorr_proof
         })
     }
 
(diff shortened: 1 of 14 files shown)


src/plonk.rs
```diff
@@ -130,9 +131,29 @@ impl<C: HaloCurve> Circuit<C> {
...
+        if cfg!(debug_assertions) {
+            // Check that division was performed correctly by evaluating at a random point.
+            let xxx = C::ScalarField::rand();
+            assert_eq!(
+                eval_poly(&plonk_t_coeffs, xxx),
+                eval_poly(&vanishing_coeffs, xxx) / eval_zero_poly(self.degree(), xxx)
+            );
+        }
 
+        // Pad the coeffiecients to length a multiple of the degree.
```
dlubarov, line 153: Hm, I thought it would always have degree `QUOTIENT_POLYNOMIAL_DEGREE_MULTIPLIER * self.degree()`. Is it the zero-trimming in `divide_by_z_h` that makes that not the case?

If it isn't already guaranteed, I think technically we should pad to that degree in particular, since the recursive circuit expects `QUOTIENT_POLYNOMIAL_DEGREE_MULTIPLIER` chunks. (Though maybe it wouldn't come up in practice.)

wborgeaud: Yes `divide_by_z_h` removes all trailing zero coefficients.
Do you mean changing the current check to something like:
```rust
if plonk_t_coeffs.len() != QUOTIENT_POLYNOMIAL_DEGREE_MULTIPLIER * self.degree() {
            plonk_t_coeffs.extend((0..QUOTIENT_POLYNOMIAL_DEGREE_MULTIPLIER * self.degree()-plonk_t_coeffs.len()).map(|_| C::ScalarField::ZERO));
        }
```
?

dlubarov: Yeah I think that makes sense, or to tweak it slightly, the range could be `plonk_t_coeffs.len()..QUOTIENT_POLYNOMIAL_DEGREE_MULTIPLIER * self.degree()`

I suspect the prover may be a tad faster if we skip zero-trimming, since these polynomials would be pretty random looking, so we'd probably never trim down to a smaller power of two (except maybe for trivial circuits). Not sure if the time used for trimming+padding is significant; I could do a comparison after this PR lands.

src/plonk_challenger.rs
```diff
@@ -22,27 +29,42 @@ impl<F: Field> Challenger<F> {
         }
     }
 
-    pub(crate) fn observe_affine_point<C: Curve<BaseField = F>>(&mut self, point: AffinePoint<C>) {
-        debug_assert!(!point.zero);
-        self.observe_element(point.x);
-        self.observe_element(point.y);
+    pub(crate) fn observe_affine_point<C: Curve<BaseField=F>>(&mut self, point: AffinePoint<C>) {
```
dlubarov, line 32: Your version seems like the natural thing to do, but it seems tricky to implement in the recursive version -- currently `AffinePointTarget` can't represent zero, and the curve-related gates can't handle zero.

Was there a case where zero was passed? If it was something like a Pedersen hash of the zero polynomial, I think we can work around that by adding a (constant) random point, maybe something hash-to-point(0) (we do that for `filler` in `curve_msm` on master).

wborgeaud: Yes I had a test that was failing because of this `assert`, but I can't remember which one, and now they all pass it. So I'll change it back like it was before.

src/plonk.rs
```diff
@@ -198,16 +219,18 @@ impl<C: HaloCurve> Circuit<C> {
...
             })
             .collect();
 
         // Generate random v, u, and x from the transcript.
         challenger.observe_elements(&all_opened_values_bf);
-        let (v_bf, u_bf, _x) = challenger.get_3_challenges();
+        let (v_bf, u_bf) = challenger.get_2_challenges();
         let v_sf = v_bf.try_convert::<C::ScalarField>()?;
         let u_sf = u_bf.try_convert::<C::ScalarField>()?;
+        let u_curve = challenger.get_affine_point::<C>();
```
dlubarov, line 233: Can this be a random constant, maybe using hash-to-point (with a constant input, or `degree` plus a constant)? That was my understanding from the paper.

wborgeaud: So, I think Bulletproof has the verifier sending a random scalar challenge, while Halo has the verifier sending `U` a random curve point. I didn't know if there were security issues with the prover knowing the discrete log of `U`, so I went with the Halo way. If you think it's safe I can change it back.

dlubarov: Hm my understanding is that the Halo paper assumes that Bulletproofs' "Protocol 1" reduction already occurred, so I think the `U` used in Halo's Schnorr protocol should be a random constant point scaled by a challenge.

src/plonk_util.rs
```diff
@@ -199,15 +209,18 @@ pub fn permutation_polynomial<F: Field>(
     gamma: F,
 ) -> Vec<F> {
     let mut plonk_z_points = vec![F::ONE];
+    let k_is = (0..NUM_ROUTED_WIRES)
+        .map(|j| get_subgroup_shift::<F>(j))
+        .collect::<Vec<_>>();
     for i in 1..degree {
-        let x = subgroup[i];
+        let x = subgroup[i - 1];
```
dlubarov, line 216: I think the recursive `S_id` / `S_sigma` code in `verify_assumptions` may need to be tweaked to match this.

src/plonk_util.rs
```diff
@@ -216,3 +229,101 @@ pub fn permutation_polynomial<F: Field>(
...
+pub fn sigma_polynomials<F: Field>(
+    sigma: Vec<usize>,
+    degree: usize,
+    subgroup_generator: F,
+) -> Vec<Vec<F>> {
+    sigma
+        .chunks(degree)
+        .map(|chunk| {
+            chunk
+                .iter()
```
dlubarov, line 242: Could you switch this to par_iter()? I'm guessing the field exps below would take a nontrivial amount of time, but probably small enough if it's parallelized.

src/verifier.rs
```diff
@@ -0,0 +1,406 @@
...
+use crate::plonk_proof::OpeningSet;
+use crate::plonk_util::{eval_poly, halo_g, halo_n, pedersen_hash, powers, reduce_with_powers};
+
+const SECURITY_BITS: usize = 128;
+
+pub struct VerificationKey<C: Curve> {
+    selector_commitments: Vec<AffinePoint<C>>,
+    sigma_commitments: Vec<AffinePoint<C>>,
+    degree_log: usize,
+    degree_pow: usize,
```
dlubarov, line 22: Oh I've been using `degree_pow` as the log, but `degree_log` does seem more clear, so let's use `degree` & `degree_log`?

src/verifier.rs
```diff
@@ -0,0 +1,406 @@
...
+}
+
+pub fn verify_proof_circuit<C: HaloCurve, InnerC: HaloCurve<BaseField=C::ScalarField>>(
+    public_inputs: &[C::ScalarField],
+    proof: &Proof<C>,
+    circuit: &Circuit<C>,
+) -> Result<bool> {
```
dlubarov, line 29: Maybe we should just use Result<()> so every failure would return an error string? To me it's not obvious what type of error `Ok(false)` would signify.

src/verifier.rs
```diff
@@ -0,0 +1,406 @@
...
+
+pub struct VerificationKey<C: Curve> {
+    selector_commitments: Vec<AffinePoint<C>>,
+    sigma_commitments: Vec<AffinePoint<C>>,
+    degree_log: usize,
+    degree_pow: usize,
+}
+
+pub fn verify_proof_circuit<C: HaloCurve, InnerC: HaloCurve<BaseField=C::ScalarField>>(
+    public_inputs: &[C::ScalarField],
```
dlubarov, line 26: Could we extract public inputs from the proof, something like this?

    (0..circuit.num_public_inputs).map(|i| proof.o_public_inputs[i / NUM_WIRES].o_wires[i % NUM_WIRES]).collect()

I see you have similar logic in `verify_public_inputs`; I think it would just be slightly nicer if users had one less argument to pass.

src/verifier.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    let constraint_terms = evaluate_all_constraints::<C, InnerC>(
+        &proof.o_local.o_constants,
+        &proof.o_local.o_wires,
+        &proof.o_right.o_wires,
+        &proof.o_below.o_wires,
+    );
+
+    // Evaluate zeta^degree.
+    let mut zeta_power_d = challs.zeta.exp_usize(degree);
```
dlubarov, line 65: Nit: `mut` not needed

src/verifier.rs
```diff
@@ -0,0 +1,406 @@
...
+    // Evaluate Z_H(zeta).
+    let one = <C::ScalarField as Field>::ONE;
+    let z_of_zeta = zeta_power_d - one;
```
dlubarov, line 68: nit: Would prefer `zero_of_zeta` or similar to avoid confusion with the other Z

src/plonk.rs
wborgeaud: I don't think Halo assumes Protocol 1 of Bulletproof. I think the first paragraph of Section 3.1 in the Halo paper is analogous to Protocol 1:
"The protocol takes the polynomial commitment `P`, point `x`, and claimed evaluation `v` as common inputs. In the first move the verifier sends a **random group element `U ∈ G`**. Both parties compute `P' = P + [v]U`".
Now the question is whether it matters if the prover knows the discrete log of `U`. Probably not, but to be sure I used the version where the prover doesn't know it.

dlubarov: Oh I missed that bit. Will think about that some more.

src/verifier.rs
wborgeaud: Yes you're right, went with `zero_of_zeta`.

src/verifier.rs
wborgeaud: I'm not sure about this one. I think it can be good to let the user provide the public inputs, and then verify that they are coherent with the proof's openings. Otherwise, it might be too easy to "fool" someone by providing malicious public inputs and hoping no one bothers to check them.

src/verifier.rs
wborgeaud: Yes you're right, I'll change it to `Result<()>`.

src/plonk_util.rs
wborgeaud: After further inspection, I'm not totally sure changes are needed to the `verify_assumptions` function, since all the changes I made are taken care of in the openings of the `z` and `sigma` polynomials.

PR #39, opened 2020-06-15
Minor fixes on the verifier branch

wborgeaud:


src/plonk.rs
```diff
@@ -360,7 +359,7 @@ impl<C: HaloCurve> Circuit<C> {
 
         debug_assert_eq!(halo_a.len(), 1);
         debug_assert_eq!(halo_b.len(), 1);
-        let schnorr_proof = self.schnorr_protocol(halo_a[0], halo_b[0], halo_g, randomness, u_curve, &mut challenger);
+        let schnorr_proof = self.schnorr_protocol(halo_a[0], halo_b[0], halo_g, randomness, self.u, &mut challenger);
```
dlubarov, line 362: I think `self.u` should be scaled by a random challenge to match the Bulletproofs protocol

PR #41, opened 2020-06-18
Working verifier

wborgeaud:
Working verifier with verification key support.


src/circuit_builder.rs
```diff
@@ -1258,15 +1261,25 @@ impl<C: HaloCurve> CircuitBuilder<C> {
         let constants_coeffs: Vec<Vec<C::ScalarField>> =
             values_to_coeffs(&wire_constants, &fft_precomputation_n);
         let constants_8n = coeffs_to_values_padded(&constants_coeffs, &fft_precomputation_8n);
-        let c_constants = coeffs_to_commitments(&constants_coeffs, &pedersen_g_msm_precomputation);
+        let c_constants = PolynomialCommitment::coeffs_vec_to_commitments(
+            &constants_coeffs,
+            &pedersen_g_msm_precomputation,
+            pedersen_h,
+            true,
```
dlubarov, line 1268: This can be false right (and below for c_s_sigmas)? It probably doesn't matter for performance, but it might be best if our circuit-building code is deterministic.

src/poly_commit.rs
```diff
@@ -1,144 +1,92 @@
...
+        msm_precomputation: &MsmPrecomputation<C>,
+        blinding_point: AffinePoint<C>,
+        blinding: bool,
+    ) -> Vec<Self> {
+        let mut comms: Vec<_> = coefficients_vec
+            .iter()
+            .map(|coeffs| {
+                Self::coeffs_to_commitment(coeffs, &msm_precomputation, blinding_point, blinding)
+            })
+            .collect::<Vec<_>>();
```
dlubarov, line 54: nit: don't need both type hints (either on collect or on the variable)

src/plonk.rs
```diff
@@ -95,10 +90,16 @@ impl<C: HaloCurve> Circuit<C> {
         let wire_values_8n = coeffs_to_values_padded(&wires_coeffs, &self.fft_precomputation_8n);
 
         // Commit to the wire polynomials.
-        let c_wires = coeffs_to_commitments(&wires_coeffs, &self.pedersen_g_msm_precomputation);
+        let c_wires = PolynomialCommitment::coeffs_vec_to_commitments(
+            &wires_coeffs,
+            &self.pedersen_g_msm_precomputation,
+            self.pedersen_h,
+            true,
```
dlubarov, line 97: Should this be `blinding_commitments`?

src/plonk.rs
```diff
@@ -86,6 +80,7 @@ impl<C: HaloCurve> Circuit<C> {
     pub fn generate_proof<InnerC: HaloCurve<BaseField = C::ScalarField>>(
         &self,
         witness: Witness<C::ScalarField>,
+        blinding_commitments: bool,
```
dlubarov, line 83: The verifier should also receive this flag right? If so I think it may be best as a `CircuitBuilder` config method (like `builder.set_blinding(true/false)`), so the user configures it only in one place

wborgeaud: Hm why would the verifier need to know if the commitments are blinded?

src/circuit_builder.rs
dlubarov: I think always-false seems good -- I figure if someone wants to mask what circuit they're using, it'd probably be better to commit to the VK (like in Zexe) rather than using a blinded VK, so that each commit can use different randomness.

src/plonk.rs
```diff
@@ -276,29 +287,48 @@ impl<C: HaloCurve> Circuit<C> {
...
-
         // Final IPA proof.
         let mut halo_a = reduced_coeffs;
-        let mut halo_b = powers(zeta_sf, self.degree());
+        // The Halo b vector is a random combination of the powers of all opening points.
+        let mut halo_b = self.build_halo_b(
+            &[
+                (0..2 * num_public_input_gates)
+                    .step_by(2)
+                    .map(|i| self.subgroup_generator_n.exp_usize(i))
```
dlubarov, line 300: Probably not a significant cost, but this could also be `self.subgroup_n[i]` just to avoid the exp

src/plonk.rs
dlubarov: Oh I was imagining if blinding was off, we could skip that Schnorr-based protocol in Halo (and just have the prover send `a'_0` directly), but maybe that's not significant and not worth optimizing.

How about on the prover's side -- would the benefit of disabling blinding commitments be saving some `[r] H` multiplications, or is there more to it?

I think the way you have it seems reasonable actually, just want to make sure I understand the intention.

src/plonk.rs
```diff
@@ -363,20 +391,27 @@ impl<C: HaloCurve> Circuit<C> {
...
+            randomness,
+            u_curve,
+            &mut challenger,
+        );
 
         Ok(Proof {
-            c_wires,
-            c_plonk_z,
-            c_plonk_t,
+            c_wires: c_wires.iter().map(|c| c.to_affine()).collect(),
```
dlubarov, line 404: Could also use `ProjectivePoint::batch_to_affine` (and for `c_plonk_t` below)

src/plonk_challenger.rs
```diff
@@ -30,7 +30,7 @@ impl<F: Field> Challenger<F> {
     }
 
     pub(crate) fn observe_affine_point<C: Curve<BaseField=F>>(&mut self, point: AffinePoint<C>) {
-        debug_assert!(!point.zero);
+        // debug_assert!(!point.zero);
```
dlubarov, line 33: Please add a TODO to re-enable later

src/plonk_recursion.rs
```diff
@@ -101,15 +109,18 @@ pub fn recursive_verification_circuit<C: HaloCurve, InnerC: HaloCurve<BaseField=
...
     challenger.observe_affine_point(proof.c_plonk_z);
     let alpha = challenger.get_challenge(&mut builder);
     challenger.observe_affine_points(&proof.c_plonk_t);
     let zeta = challenger.get_challenge(&mut builder);
     challenger.observe_elements(&proof.all_opening_targets());
-    let (v, u, x) = challenger.get_3_challenges(&mut builder);
+    let (v, u) = challenger.get_2_challenges(&mut builder);
+
+    // Challenge for scaling u used in the IPA verification.
+    let u_scaling = challenger.get_challenge(&mut builder);
```
dlubarov, line 123: Since we don't have prover messages between (v, u) and u_scaling, I think it would be best to keep using `get_3_challenges` just for performance (i.e. our current sponge width is 4, so we can get 3 outputs with a single permutation).

dlubarov: Or we could keep it like this and modify Challenger to store an output buffer (and discard it if the transcript changes). Maybe that would be better since it looks like `schnorr_challenge` could also be combined with `l_challenge` below, but that might make the code awkward. I could work on that later.

src/plonk_recursion.rs
```diff
@@ -170,67 +220,147 @@ fn verify_all_ipas<C: HaloCurve, InnerC: HaloCurve<BaseField=C::ScalarField>>(
...
     // Then, we reduce the above opening set reductions to a single value.
     let reduced_opening = reduce_with_powers_recursive(builder, &opening_set_reductions, v);
 
-    verify_ipa::<C, InnerC>(builder, proof, c_reduction, reduced_opening, x, ipa_challenges, security_bits)
+    let generator_n = builder.constant_wire(C::ScalarField::primitive_root_of_unity(
+        ipa_challenges.len(),
+    ));
+    let mut points: Vec<Target> = (0..2 * num_public_input_gates)
+        .step_by(2)
+        .map(|i| builder.exp_constant_usize(generator_n, i))
```
dlubarov, line 255: I think the cost of these exps will be nontrivial since we have quite a few public inputs. Could you call `constant_wire` for each point (should be significantly cheaper than the exp)? Or add a TODO for later.

dlubarov: Alternatively we could have methods like `exp_constant_usize` detect constant inputs and handle those by calling `constant_wire`. I could work on that later.

src/plonk_recursion.rs
```diff
@@ -170,67 +220,147 @@ fn verify_all_ipas<C: HaloCurve, InnerC: HaloCurve<BaseField=C::ScalarField>>(
...
-    let u = builder.constant_affine_point(
-        hash_usize_to_curve::<InnerC>((1 << ipa_challenges.len()) + 1 , security_bits)
-    );
-    let u_prime = builder.curve_mul_endo::<InnerC>(CurveMulOp { scalar: x, point: u }).mul_result;
+    // u is set to H(degree + 1).
+    let u = builder.constant_affine_point(hash_usize_to_curve::<InnerC>(
+        (1 << ipa_challenges.len()) + 1,
+        security_bits,
+    ));
+    let u_prime = builder
```
dlubarov, line 306: Oh yeah, seems like the recursive and non-recursive code diverges here, since the recursive code computes u' with the endo. Maybe add a TODO to fix later?

src/plonk_recursion.rs
```diff
@@ -170,67 +220,147 @@ fn verify_all_ipas<C: HaloCurve, InnerC: HaloCurve<BaseField=C::ScalarField>>(
...
+    });
+    let p_prime = builder.curve_add::<InnerC>(commitment, u_n_x_c);
 
     // Compute Q as defined in the Halo paper.
     let mut q_muls = Vec::new();
     for i in 0..proof.halo_l_i.len() {
         let l_i = proof.halo_l_i[i];
         let r_i = proof.halo_r_i[i];
-        let l_challenge = ipa_challenges[i];
+        let l_challenge = builder.square(ipa_challenges[i]);
```
dlubarov, line 325: This doesn't hurt (assuming the prover & other verifier match), but these scalars will be passed through that injective `n` before we take their square roots, so I don't think squaring at this point guarantees that we'll be able to take the square root later.

I haven't gotten around to implementing it yet, but this was my plant to make sure the sqrt below doesn't fail, after discussing a bit with Daira in that Telegram group: when we sample each IPA challenge `x`, we could check if `n(x)` is square, and if not, rewind and sample new randomness for `L_j` / `R_j`. (Challenger doesn't support rewinding currently, but we can make it copy-on-write or something.)

src/plonk_recursion.rs
```diff
@@ -170,67 +220,147 @@ fn verify_all_ipas<C: HaloCurve, InnerC: HaloCurve<BaseField=C::ScalarField>>(
...
+        .step_by(2)
+        .map(|i| builder.exp_constant_usize(generator_n, i))
+        .collect();
+    points.extend(&[zeta, builder.mul(zeta, generator_n), {
+        let g65 = builder.exp_constant_usize(generator_n, GRID_WIDTH);
+        builder.mul(zeta, g65)
+    }]);
+    let halo_bs = points
+        .into_iter()
+        .map(|p| halo_g_recursive(builder, p, &ipa_challenges))
```
dlubarov, line 263: In my mind `ipa_challenges` has the "raw" challenges (i.e. the randomness the verifier samples), but it's different from the `u` values `g` expects because of the endo. It might be a significant amount of work to fix all this stuff related to the IPA challenges, but could you add a TODO to note that this isn't the data `g` expects?

src/plonk_util.rs
```diff
@@ -257,6 +246,36 @@ pub(crate) fn polynomial_degree<F: Field>(
     coeffs.iter().rev().skip_while(|c| c.is_zero()).count() - 1
 }
 
+pub fn build_s<F: Field>(us: &[F]) -> Vec<F> {
```
dlubarov, line 249: Could you either call it `halo_s`, or add a comment to mention that it's `s` in the context of the Halo paper?

src/verifier.rs
```diff
@@ -223,23 +306,52 @@ fn verify_all_ipas<C: HaloCurve>(
     // Then, we reduce the above opening set reductions to a single value.
     let reduced_opening = reduce_with_powers(&opening_set_reductions, v);
 
-    let u_curve = C::convert(u_scaling) * circuit.u.to_projective();
-
-    verify_ipa::<C>(proof, c_reduction, reduced_opening, point, ipa_challenges, u_curve, circuit.pedersen_h,proof.halo_g, schnorr_challenge, proof.schnorr_proof)
+    let u_curve = C::convert(u_scaling) * u_curve.to_projective();
```
dlubarov, line 309: I'd prefer if the variable names differed -- maybe call the new one u_prime or u_scaled?

tests/prove_and_verify.rs
```diff
@@ -0,0 +1,151 @@
...
+    let witness = circuit.generate_witness(partial_witness);
+    let proof = circuit.generate_proof::<Tweedledum>(witness, true).unwrap();
+    assert!(verify_proof_circuit::<Tweedledee, Tweedledum>(&[], &proof, &circuit,).is_ok());
+}
+
+#[test]
+fn test_proof_sum_big_circuit() {
+    let now = Instant::now();
+    let mut builder = CircuitBuilder::<Tweedledee>::new(128);
+    let ts = (0..10_000)
```
dlubarov, line 40: Don't need to change anything now, but I think it would be nice if `cargo test` only ran fast tests by default, so we could run that regularly. Not sure how best to handle large tests -- we could `#[ignore]` them, or use a cfg rule, or make them binaries.

src/plonk.rs
wborgeaud: Yes without blinding we can skip the Schnorr protocol and spare a few operations. 
The way I see it, we would turn off blinding only for inner proofs in a recursive argument, where only the last step needs blinding. In particular, the out-of-circuit verifier would practically never be run on a proof without blinding.
The in-circuit verifier however should practically be run only on proof without blinding. So I could remove the Schnorr protocol from `recursive_verification_circuit` in this case.

src/plonk.rs
wborgeaud: Actually in this case `c_wires` and `c_plonk_t` are  `PolynomialCommitment` structs, and I assume that `PolynomialCommitment::batch_to_affine` has been run on them previously, so the `to_affine` is basically costless.
The `PolynomialCommitment` is a bit wonky so I can change it if you want.

src/plonk_recursion.rs
wborgeaud: I'll change it to `get_3_challenges` for now.

src/plonk_recursion.rs
wborgeaud: I'll call `constant_wires` for now.

src/plonk_recursion.rs
wborgeaud: Ah yes you're right. What do you think is best here, changing this one to `curve_mul` or changing the out-of-circuit stuff to use the endo scalar multiplication?

src/plonk_recursion.rs
wborgeaud: I added a TODO for this.
It seems like the loop to check if `n(x)` is square and backtrack if not will be expensive. Maybe there is a way to define another function that maps to squares instead.

src/plonk_recursion.rs
wborgeaud: Couldn't we store `n(u)` in `ipa_challenges` instead?

wborgeaud: By the way, I think this
```rust
halo_a = C::ScalarField::add_slices(
    &l_challenge_sf.scale_slice(a_lo),
    &r_challenge_sf.scale_slice(a_hi),
);
halo_b = C::ScalarField::add_slices(
    &l_challenge_sf.scale_slice(b_hi),
    &r_challenge_sf.scale_slice(b_lo),
);
```
in the prover (`plonk.rs` 371-378`) is wrong too, and actually all the proof/verification uses the "wrong" challenges. If we just apply `n` before storing in `ipa_challenges`, it should work though.

src/plonk_util.rs
wborgeaud: I changed the name and remove the `pub` since I only use it to test the `g` function.

tests/prove_and_verify.rs
wborgeaud: I `#[ignore]`'d all the big tests.

src/plonk_recursion.rs
wborgeaud: Also, why can't we use `n(x)` and `n(x)^2` instead of `√n(x)` and `n(x)`?

src/plonk.rs
dlubarov: Hm I think if we consider use cases outside of Mir, there could be potential uses for all the modes -- like if someone was (out-of-circuit) verifying a SNARK-based VDF, that could be non-blinded since there's no privacy requirement, or if a block producer was recursing on the previous block proof, that might be blinded if the inner proof was generated by a different party. So I think ideally we should support all modes, but it's probably a minor cost savings so feel free to leave it for later.

src/plonk.rs
dlubarov: Oh I see, looks reasonable then.

src/plonk_recursion.rs
dlubarov: Yeah I think the recursive verifier is enforcing that each `n(ipa_challenge[i])` is square, since IIRC it calls `deterministic_square_root` on those values, and that method checks the square of the purported sqrt. (The verifier isn't computing `n` explicitly, but gets those values implicitly from CurveEndoGate's "signed accumulator" wires.)

PR #42, opened 2020-06-19
Improve halo_n and u_prime

wborgeaud:
From this discussion: https://github.com/mir-protocol/plonky/pull/41#discussion_r442426267
Fixed of the definition of `u_prime` to include scaling with endo. I also wrote the implementation of the paper's Algorithm 1 out-of-circuit.


src/curve.rs
```diff
@@ -44,6 +44,18 @@ pub trait Curve: 'static + Sync + Sized + Copy + Debug {
 pub trait HaloCurve: Curve {
     const ZETA: Self::BaseField;
     const ZETA_SCALAR: Self::ScalarField;
+
+    fn endomorphism(p: AffinePoint<Self>) -> AffinePoint<Self> {
```
dlubarov, line 48: Maybe have it take `&self`, or do you think `p.endomorphism()` doesn't look as natural?

src/field.rs
```diff
@@ -130,7 +130,12 @@ pub trait Field:
     }
 
     fn from_canonical_bool(b: bool) -> Self {
-        Self::from_canonical_u64(if b { 1 } else { 0 })
+        if b {
+            Self::ONE
+        } else {
+            Self::ZERO
+        }
+        // Self::from_canonical_u64(if b { 1 } else { 0 })
```
dlubarov, line 138: I like your version; we can delete this

src/plonk_util.rs
```diff
@@ -77,6 +74,44 @@ pub(crate) fn halo_n<C: HaloCurve>(s_bits: &[bool]) -> C::ScalarField {
     a * C::ZETA_SCALAR + b
 }
 
+/// Compute `n(x)` for a given `x`, where `n` is the injective function related to the Halo
```
dlubarov, line 77: Seems slightly out of date -- let's say "Compute `[n(s)] P`"

src/plonk_util.rs
```diff
@@ -77,6 +74,44 @@ pub(crate) fn halo_n<C: HaloCurve>(s_bits: &[bool]) -> C::ScalarField {
...
+/// endomorphism.
+pub(crate) fn halo_n_mul<C: HaloCurve>(s_bits: &[bool], p: AffinePoint<C>) -> AffinePoint<C> {
```
dlubarov, line 79: nit: maybe call it halo_endo_mul? Just seems easy to forget what `n` refers to in Halo

src/plonk_util.rs
```diff
@@ -77,6 +74,44 @@ pub(crate) fn halo_n<C: HaloCurve>(s_bits: &[bool]) -> C::ScalarField {
...
+    let endo_p_p = C::endomorphism(p).to_projective();
+    let endo_p_n = endo_p_p.neg();
+
+    let mut acc = ProjectivePoint::<C>::ZERO;
+
+    for s_bits_chunk in s_bits.chunks(2) {
+        let bit_lo = s_bits_chunk[0];
+        let bit_hi = s_bits_chunk[1];
+
+        let s;
```
dlubarov, line 95: nit: maybe slightly nicer as `let s = if bit_hi { ... } else { ... }`, or do you not like that style?

src/plonk_util.rs
```diff
@@ -77,6 +74,44 @@ pub(crate) fn halo_n<C: HaloCurve>(s_bits: &[bool]) -> C::ScalarField {
...
+        if bit_hi {
```
dlubarov, line 96: I think this is switched from the Halo paper, where the first branch is a hi bit of 0. Doesn't really matter but I think I used that convention in the recursive version.

src/plonk_util.rs
```diff
@@ -77,6 +74,44 @@ pub(crate) fn halo_n<C: HaloCurve>(s_bits: &[bool]) -> C::ScalarField {
...
+                s = p_n;
+            }
+        } else {
+            if bit_lo {
+                s = endo_p_p;
+            } else {
+                s = endo_p_n;
+            }
+        }
+        acc = acc + acc + s;
```
dlubarov, line 109: Not very important but `acc.double()` should be a bit faster

src/curve.rs
```diff
@@ -175,19 +205,46 @@ impl<C: Curve> ProjectivePoint<C> {
...
     pub fn add_slices(a: &[Self], b: &[Self]) -> Vec<Self> {
         assert_eq!(a.len(), b.len());
-        a.iter().zip(b.iter()).map(|(&a_i, &b_i)| a_i + b_i).collect()
+        a.iter()
+            .zip(b.iter())
+            .map(|(&a_i, &b_i)| a_i + b_i)
+            .collect()
+    }
+
+    pub fn neg(&self) -> Self {
```
dlubarov, line 224: There's also an `impl Neg` for AffinePoint & ProjectivePoint, so `-p` should work

src/plonk_util.rs
dlubarov: On second thought, since we do have a `halo_n` method, maybe it's good to match that. Up to you.

wborgeaud: I'll think I'll keep `halo_n_mul` for now to be coherent with the `halo_n` function.

src/plonk_util.rs
wborgeaud: Yes you're right. It was switched in the `halo_n` function, so I've changed both `halo_n` and `halo_n_mul` and it should be coherent with the paper and `CurveEndoGate` now.

PR #43, opened 2020-06-26
Add bin for running MSMs with various thread pool sizes

dlubarov:


dlubarov:
With this, we'd run the old bin with

    cargo run --bin recursion --release

And this other one with

    cargo run --bin msms --release

dlubarov:
Also added a default so `cargo run --release` still works (defaults to recursion)

PR #45, opened 2020-06-29
Fix curve_add

dlubarov:
It was just missing some routing.

### DIFF
--- a/src/circuit_builder.rs
+++ b/src/circuit_builder.rs
@@ -695,6 +695,7 @@ impl<C: HaloCurve> CircuitBuilder<C> {
         p_1: AffinePointTarget,
         p_2: AffinePointTarget,
     ) -> AffinePointTarget {
+        // Add a CurveAddGate, then add a BufferGate to receive the updated accumulator state.
         let add_index = self.num_gates();
         self.add_gate_no_constants(CurveAddGate::<C, InnerC>::new(add_index));
         let buffer_index = self.num_gates();
--- a/src/circuit_builder.rs
+++ b/src/circuit_builder.rs
@@ -692,16 +692,38 @@ impl<C: HaloCurve> CircuitBuilder<C> {
 
     pub fn curve_add<InnerC: Curve<BaseField = C::ScalarField>>(
         &mut self,
-        _p_1: AffinePointTarget,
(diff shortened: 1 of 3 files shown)


wborgeaud:
I tried to test `curve_add` as follows:
```rust
#[test]
fn test_curve_add() {
    type F = <Tweedledee as Curve>::ScalarField;

    let a = blake_hash_base_field_to_curve::<Tweedledum>(F::rand());
    let b = blake_hash_base_field_to_curve::<Tweedledum>(F::rand());
    let sum = (a + b).to_affine();

    let mut builder = CircuitBuilder::<Tweedledee>::new(128);

    let ta =  builder.add_virtual_point_target();
    let tb =  builder.add_virtual_point_target();
    let tsum_purported = builder.curve_add::<Tweedledum>(ta, tb);
    let tsum_true = builder.constant_affine_point(sum);
    let should_be_zero = builder.curve_sub::<Tweedledum>(tsum_purported, tsum_true);

    builder.assert_zero(should_be_zero.x);
    builder.assert_zero(should_be_zero.y);

    let mut partial_witness = PartialWitness::new();
    partial_witness.set_point_target(ta, a);
    partial_witness.set_point_target(tb, b);

    let circuit = builder.build();
    let witness = circuit.generate_witness(partial_witness);

    let proof = circuit.generate_proof::<Tweedledum>(witness, true).unwrap();
    assert!(verify_proof_circuit::<Tweedledee, Tweedledum>(&[], &proof, &circuit,).is_ok());
}
```
but I get a panic `Non-routed wires should not be in a partition containing other wires', src/partition.rs:89:21`. Is there a problem with my test?

wborgeaud:
I wrote a fix here https://github.com/mir-protocol/plonky/commit/9a85826321973d3f7b011ccacdd5d22d8e7049f4. I had to basically remove the `scalar_bit` and set it to one by default. The problem is that it's at index 6 and should thus be non-routable. 
I don't know if that's an acceptable fix for our applications.

dlubarov:
Sorry I forgot about this, taking a look now.

wborgeaud:
Actually I just saw that you use the scalar bit for the MSM, so my fix doesn't work.

PR #46, opened 2020-07-03
Various fixes

wborgeaud:
- Add "old proofs" logic.
- Add full verification of the IPA argument.
- Various bug fixes.


src/bin/recursion.rs
```diff
@@ -1,12 +1,13 @@
...
 
-use plonky::{recursive_verification_circuit, verify_proof_circuit, BufferGate, Circuit, CircuitBuilder, PartialWitness, Tweedledee, Tweedledum};
+use plonky::{recursive_verification_circuit, verify_proof_circuit, BufferGate, Circuit, CircuitBuilder, Field, PartialWitness, RecursionPublicInputs, Tweedledee, Tweedledum};
 
 const INNER_PROOF_DEGREE_POW: usize = 14;
 const INNER_PROOF_DEGREE: usize = 1 << INNER_PROOF_DEGREE_POW;
 const SECURITY_BITS: usize = 128;
 
-fn main() {
+fn main() -> Result<()> {
```
dlubarov, line 10: Didn't know about that option, that's nice!

src/bin/recursion.rs
```diff
@@ -21,25 +22,34 @@ fn main() {
...
         .unwrap();
     println!("Finished in {}s", start.elapsed().as_secs_f64());
     println!();
 
     println!("Verifying inner proof");
     let start = Instant::now();
-    assert!(
-        verify_proof_circuit::<Tweedledum, Tweedledee>(&[], &inner_proof, &inner_circuit).is_ok()
-    );
+    assert!(verify_proof_circuit::<Tweedledum, Tweedledee>(
```
dlubarov, line 34: Can this use `?` now?

src/bin/recursion.rs
```diff
@@ -64,19 +88,26 @@ fn main() {
...
         .unwrap();
     println!("Finished in {}s", start.elapsed().as_secs_f64());
     println!();
 
     println!("Verifying proof...");
     let start = Instant::now();
+    let pis =
+        RecursionPublicInputs::proof_to_public_inputs::<Tweedledum, Tweedledee>(&inner_proof, &[])?;
+    dbg!(recursion_circuit.circuit.num_public_inputs);
     dbg!(verify_proof_circuit::<Tweedledee, Tweedledum>(
```
dlubarov, line 102: Can this use `?` now?

src/witness.rs
```diff
@@ -85,6 +85,11 @@ impl<F: Field> PartialWitness<F> {
         self.set_target(Target::Wire(wire), value);
     }
 
+    pub fn set_public_input(&mut self, pi: PublicInput, value: F) {
+        self.set_wire(pi.original_wire(), value);
+        self.set_target(pi.routable_target(), value);
```
dlubarov, line 90: It should work without this since `PublicInputGate`'s generator copies the original wire to the routable wire, but also OK to keep if you prefer.

src/bin/recursion.rs
wborgeaud: Yes it's useful. It also works with tests by the way.

src/witness.rs
wborgeaud: Oh yeah, I think I added this to fix a bug, but you're right it's useless now 👍

src/gates/curve_add.rs
```diff
@@ -204,7 +196,7 @@ for CurveAddGate<C, InnerC>
         let addend_x = witness.get_wire(addend_x_target);
         let addend_y = witness.get_wire(addend_y_target);
 
-        let scalar_bit = witness.get_wire(scalar_bit_target);
+        let scalar_bit = C::ScalarField::ONE;
```
dlubarov, line 199: Think this should be changed back

src/plonk_proof.rs
```diff
@@ -133,6 +254,16 @@ impl ProofTarget {
         witness.set_point_targets(&self.halo_r_i, &values.halo_r);
         witness.set_point_target(self.halo_g, values.halo_g);
 
+        witness.set_point_target(self.schnorr_proof.r, values.schnorr_proof.r);
+        witness.set_target(
+            self.schnorr_proof.z1,
+            C::ScalarField::try_convert(&values.schnorr_proof.z1)?,
+        );
+        witness.set_target(
+            self.schnorr_proof.z1,
```
dlubarov, line 263: Should be z2?

src/plonk_recursion.rs
```diff
@@ -47,9 +48,100 @@ impl RecursionPublicInputs {
             self.o_plonk_t.as_slice(),
             self.halo_u_l.as_slice(),
             self.halo_u_r.as_slice(),
+            self.old_proofs.as_slice(),
         ]
         .concat()
     }
+
+    pub fn populate_witness<C: Curve>(
```
dlubarov, line 56: Hmm my expectation was that callers would populate the `ProofTarget` and everything else would be generated by the circuit. The code near the end of `recursive_verification_circuit` should copy claimed openings from `ProofTarget`'s wire to the public input wires, and the challenge elements should be generated by `Rescue*` gates. So I don't think we really need these steps, but let me know if the generators aren't working properly and I can debug.

src/plonk_recursion.rs
```diff
@@ -47,9 +48,100 @@ impl RecursionPublicInputs {
...
+                    )),
+            )
+        {
+            witness.set_public_input(a, b.try_convert()?);
+        }
+
+        Ok(())
+    }
+
+    pub fn proof_to_public_inputs<C: Curve, InnerC: HaloCurve<BaseField = C::ScalarField>>(
```
dlubarov, line 107: I think it should be possible to do this in a generic way, with public inputs being an output of `generate_proof` (maybe contained in `Proof`). I could try implementing that later.

src/plonk_recursion.rs
```diff
@@ -194,9 +294,31 @@ pub fn recursive_verification_circuit<
         builder.copy(public_inputs.halo_u_l[i].routable_target(), u_l[i]);
         builder.copy(public_inputs.halo_u_r[i].routable_target(), u_r[i]);
     }
+    let shift = 2 + degree_pow;
+    for i in 0..num_old_proofs {
+        builder.copy(
+            old_proofs[i].halo_g.x,
```
dlubarov, line 300: Hm if I'm reading this correctly, it seems like it's copying "inner old proof" data to "outer old proof" data?

I think we should copy `proof.halo_g` into and `ipa_challenges` into `public_inputs.old_proofs`. Although it's a bit awkward since we have one each of those variables for now, but long term we should have one `halo_g` and `ipa_challenges` for each inner proof, and the same number of "old proof outputs", if I understand correctly.

src/plonk_recursion.rs
```diff
@@ -558,9 +697,21 @@ fn halo_g_recursive<C: HaloCurve>(
     let mut x_power = x;
     for &u_i in us.iter().rev() {
         let u_i_inv = builder.inv(u_i);
-        let term = builder.mul_add(u_i_inv, x_power, u_i);
+        let term = builder.mul_add(u_i, x_power, u_i_inv);
```
dlubarov, line 700: Was this causing an issue? I think the old version matches the convention in the Halo paper.

src/verifier.rs
```diff
@@ -106,19 +143,28 @@ pub struct VerificationKey<C: Curve> {
     pub security_bits: usize,
 }
 
+/// Verifies a proof `proof` and some old proofs G points for a given verification key.
+/// If `verify_g` is `true`, the function completely verifies the proof, including the
+/// linear time check of the G point.
+/// If `verify_g` is `false`, returns an `OldProof` to be checked in a later verification.
 pub fn verify_proof_vk<C: HaloCurve, InnerC: HaloCurve<BaseField = C::ScalarField>>(
```
dlubarov, line 150: By the way, would `verify_proof_vk` the preferred verification method generally since the VK is smaller? Can we get rid of `verify_proof_circuit` later?

src/plonk_util.rs
```diff
@@ -280,7 +280,8 @@ pub(crate) fn polynomial_degree<F: Field>(
     coeffs.iter().rev().skip_while(|c| c.is_zero()).count() - 1
 }
 
-fn halo_s<F: Field>(us: &[F]) -> Vec<F> {
+// TODO: Maybe a streaming version using an `Iterator` would be faster and wouldn't require as much memory for large circuits.
```
dlubarov, line 283: I think we could also do this with 2*degree field muls by exploiting the common terms. Not sure if it's worth the work to optimize though.

src/plonk_recursion.rs
wborgeaud: Yes we definitely could. I don't have a strong opinion on this, so if you think it's better to have the PI as an output of `generate_proof` we can implement that in a later PR.

src/plonk_recursion.rs
wborgeaud: What do you mean by "outer old proof" data?
What I'm trying to do here is just continue copying the public inputs to the virtual targets, similar to what is done from lines 255-296, e.g.,
```rust
builder.copy(
        public_inputs.o_plonk_z_right.routable_target(),
        proof.o_right.o_plonk_z,
    );
```

src/plonk_recursion.rs
wborgeaud: So I think there's a mistake in the Halo paper and they've swapped `u_i` and `u_i^-1` in the definition of the `g` polynomial. For example, the 0-th element in the `s` vector is `prod u_i^-1`, while the 0-th degree term of `g` is `prod u_i`.

src/plonk_util.rs
wborgeaud: Yes I agree there should be a way to greatly optimize this. I'll add a TODO.

tests/prove_and_verify.rs
```diff
@@ -1,20 +1,57 @@
...
+}
+
+#[test]
+fn test_proof_trivial_circuit() {
+    let (circuit, witness) = get_trivial_circuit(<Tweedledee as Curve>::ScalarField::ZERO);
+    let proof = circuit
+        .generate_proof::<Tweedledum>(witness.clone(), &[], true)
+        .unwrap();
+    assert!(
+        verify_proof_circuit::<Tweedledee, Tweedledum>(&[], &proof, &[], &circuit, true).is_ok()
```
dlubarov, line 25: Maybe nice to use either `expect(msg)`, `unwrap()`, or `?` with a `Result` return type (as you mentioned) just so we'll see the Result's error easily upon failure.

tests/prove_and_verify.rs
```diff
@@ -147,6 +209,113 @@ fn test_proof_sponge_circuit() {
...
+    let tres_purported = builder.curve_msm::<Tweedledee>(
+        &(0..n)
+            .map(|i| CurveMulOp {
+                scalar: txs[i],
+                point: tps[i],
+            })
+            .collect::<Vec<_>>(),
+    );
+    let tres_true = builder.constant_affine_point(res.to_affine());
+    let should_be_zero = builder.curve_sub::<Tweedledee>(tres_purported, tres_true);
```
dlubarov, line 302: Does this work? I thought we'd get a witness generation error since the curve gates don't handle exceptional cases.

src/plonk_recursion.rs
dlubarov: Hm I think of those other copies as "outputting" bits of the inner proof that still need to be checked. If I understand correctly, the `g` points in `old_proofs` have been fully processed by this point, since they're incorporated into this circuit's opening argument by `verify_old_proof_evaluation`. It seems like `proof.halo_g` is the one that still needs to be checked by the next proof. Is that right?

src/plonk_recursion.rs
dlubarov: Good point!

src/plonk_recursion.rs
wborgeaud: Ok I see what you mean. I'll make the changes in another PR.

src/verifier.rs
wborgeaud: Yes that was the plan, or at least I've always thought that in applications we would use the verifying key. I'll keep the function for now as it can be useful for debugging, but later we can either delete it or wrap it in a `cfg(test) / cfg(debug)`.

tests/prove_and_verify.rs
wborgeaud: Yes you're right. Though this test fails for another reason, but `test_curve_add` panics if I use the `should_be_zero` method. I've changed all of these to two copy constraints.

PR #47, opened 2020-07-06
Fix degree of CurveAddGate

dlubarov:


src/plonk_util.rs
```diff
@@ -272,12 +272,12 @@ pub fn sigma_polynomials<F: Field>(
         .collect()
 }
 
-pub(crate) fn polynomial_degree<F: Field>(
+pub(crate) fn polynomial_degree_plus_1<F: Field>(
```
dlubarov, line 275: Does this seem reasonable? The motivation was to avoid underflow in the case where the polynomial is all zeros, which came up with one of the constraint polynomials.

wborgeaud: Yes looks good.

src/gates/curve_add.rs
```diff
@@ -36,37 +36,50 @@ impl<C: HaloCurve, InnerC: Curve<BaseField = C::ScalarField>> CurveAddGate<C, In
...
         _local_constant_values: &[InnerC::BaseField],
         local_wire_values: &[InnerC::BaseField],
         right_wire_values: &[InnerC::BaseField],
         _below_wire_values: &[InnerC::BaseField],
     ) -> Vec<InnerC::BaseField> {
+        // Notation:
+        // - p1 is the accumulator;
+        // - p2 is the addend;
+        // - p3 = p1 + p2;
+        // - p4 = if scalar_bit { p4 } else { p1 }
```
wborgeaud, line 51: Should be `p4 = if scalar_bit { p3 } else { p1 }` I think.

src/gates/curve_add.rs
```diff
@@ -36,37 +36,50 @@ impl<C: HaloCurve, InnerC: Curve<BaseField = C::ScalarField>> CurveAddGate<C, In
...
         let computed_lambda = (y1 - y2) * inverse;
-        let computed_x3 = lambda.square() - x1 - x2;
-        let computed_y3 = lambda * (x1 - x3) - y1;
+        let x3 = lambda.square() - x1 - x2;
+        // We subtract x4 instead of x3 in order to minimize degree. This will give an incorrect
+        // result for y3 if x3 != x4, which happens when scalar_bit = 0, but in that case y3 will
+        // be ignored (i.e. multiplied by zero), so we're okay.
+        let y3 = lambda * (x1 - x4) - y1;
+
+        let not_scalar_bit = scalar_bit - InnerC::BaseField::ONE;
```
wborgeaud, line 73: Should be `not_scalar_bit = InnerC::BaseField::ONE - scalar_bit` I think.

dlubarov: Good catch!

PR #49, opened 2020-07-07
Msm fixes

wborgeaud:
Fixes the `CircuitBuilder::curve_msm` function.


src/gates/curve_add.rs
```diff
@@ -232,7 +264,113 @@ impl<C: HaloCurve, InnerC: Curve<BaseField = C::ScalarField>> WitnessGenerator<C
 
 #[cfg(test)]
 mod tests {
-    use crate::{test_gate_low_degree, CurveAddGate, Tweedledee, Tweedledum};
+    use crate::{test_gate_low_degree, CurveAddGate, Field, Tweedledee, Tweedledum};
+
+    #[test]
+    // Similar to the test below, except we manually set the scalar bit wire to a random bit instead of a random field element.
+    fn test_curve_add_is_low_degree() {
```
wborgeaud, line 271: Not sure what to do with this test. It's very redundant with the macro test below, but I'd still want to keep the logic to set the scalar bit to a random bit to avoid false positives in the future.

dlubarov: Hm, I'm not sure why the false positives would happen; in theory I don't think we should need 0/1 to detect high degreeness. Maybe there's a bug in `test_gate_low_degree`? Let me know if you notice an easy way to reproduce the false positives and I can look into it.

wborgeaud: I figured out what happened. In my first iteration of rewriting CurveAddGate, I just used a dumb conditional `if scalar_bit == 1 {...}`, so I needed at least one 1 in the random `scalar_bit` to make the test fail.
Now that we use the correct version of `CurveAddGate`, I agree it's no longer necessary and I will remove this test.

dlubarov: Ah that makes sense! In a future rewrite maybe we could have `evaluate_unfiltered` return a polynomial representation rather than evaluating at a point, so that things like if/else wouldn't be possible and we could detect if the degree exceeded 8.

PR #51, opened 2020-07-07
Some fixes relating to our use of the Halo endo

dlubarov:
This should compute `halo_us` properly. (I'm avoiding "IPA challenges" now because of the ambiguity.)

The main method is now failing with a PI related error, but I think it makes sense. The recursive circuit expects its inner proof to publish assumptions as PIs, but the dummy proof has no assumptions or PIs.

### DIFF
--- a/src/circuit_builder.rs
+++ b/src/circuit_builder.rs
@@ -1177,7 +1177,7 @@ impl<C: HaloCurve> CircuitBuilder<C> {
             // Normally, we will take the double gate's outputs as the new accumulator. If we just
             // completed the last iteration of the MSM though, then we don't want to perform a final
             // doubling, so we will take its inputs as the result instead.
-            if i == scalar_bits - 1 {
+            if i == 0 {
                 acc = AffinePointTarget {
                     x: Target::Wire(Wire {
                         gate,
@@ -1199,10 +1199,10 @@ impl<C: HaloCurve> CircuitBuilder<C> {
                         input: CurveDblGate::<C, InnerC>::WIRE_Y_NEW,
                     }),
                 };
-            }
 
-            // Also double the filler, so we can subtract out a rescaled version later.
-            filler = filler.double();
+                // Also double the filler, so we can subtract out a rescaled version later.
+                filler = filler.double();
+            }
         }
 
         // Subtract (a rescaled version of) the arbitrary nonzero value that we started with.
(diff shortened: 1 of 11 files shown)


src/plonk_recursion.rs
```diff
@@ -46,63 +45,10 @@ impl RecursionPublicInputs {
...
             &[self.o_plonk_z_local, self.o_plonk_z_right],
             self.o_plonk_t.as_slice(),
-            self.halo_u_l.as_slice(),
-            self.halo_u_r.as_slice(),
             self.old_proofs.as_slice(),
         ]
         .concat()
     }
-
-    pub fn populate_witness<C: Curve>(
```
dlubarov, line 56: (I think the circuit should be generating all these values from the inputs passed to `ProofTarget::populate_witness`)

src/circuit_builder.rs
```diff
@@ -849,6 +851,56 @@ impl<C: HaloCurve> CircuitBuilder<C> {
...
+                    .try_convert::<InnerC::ScalarField>().expect("Improbable");
+                let scalar_inv = scalar.multiplicative_inverse().expect("Can't invert zero");
+                let point = witness.get_point_target(self.mul.point);
+                let result = (InnerC::convert(scalar_inv) * point.to_projective()).to_affine();
+
+                let mut result_witness = PartialWitness::new();
+                result_witness.set_point_target(self.result, result);
+                result_witness
+            }
+        }
```
wborgeaud, line 888: Doesn't this compute `[1/s] p` instead of `[1/n(s)] p`?

dlubarov: Oh yeah you're right. I fixed that but `curve_mul_inv_endo` is still failing; looking into it now.

src/plonk_recursion.rs
```diff
@@ -30,8 +30,7 @@ pub struct RecursionPublicInputs {
     o_plonk_z_local: PublicInput,
     o_plonk_z_right: PublicInput,
     o_plonk_t: Vec<PublicInput>,
-    halo_u_l: Vec<PublicInput>,
-    halo_u_r: Vec<PublicInput>,
+    halo_us: Vec<PublicInput>,
     old_proofs: Vec<PublicInput>,
 }
```
wborgeaud, line 36: Is the `to_vec` function used somewhere?

src/circuit_builder.rs
wborgeaud: I think there's a problem with `curve_msm_endo`, so I think once that's fixed `curve_mul_inv_endo` should work.

dlubarov: Agreed. It looks like it had the same two bugs that you fixed in the non-endo MSM method recently, so I added your fixes in there. It seems there another issue related to how we use the g^65 opening; looking into that now.

src/plonk_recursion.rs
dlubarov: Oh yeah seems it's not used. Deleted it :)

PR #52, opened 2020-07-10
Modify Challenger to use a duplex sponge pattern

dlubarov:
This may save us some evaluations of the permutation, since we no longer have to save one of the outputs and pass it in as an input to the next hash. The sponge already has "memory" of previous messages via its capacity element.

Inspired by @daira's talk: https://www.youtube.com/watch?v=IQVGIqcdxL4&t=50m05s

### DIFF
--- a/src/plonk_challenger.rs
+++ b/src/plonk_challenger.rs
@@ -108,12 +108,6 @@ impl<F: Field> Challenger<F> {
     /// Absorb any buffered inputs. After calling this, the input buffer will be empty.
     fn absorb_buffered_inputs(&mut self) {
         for input_chunk in self.input_buffer.chunks(RESCUE_SPONGE_RATE) {
-            // If this is a partial chunk, zero-pad it.
-            let mut input_chunk = input_chunk.to_vec();
-            while input_chunk.len() < RESCUE_SPONGE_RATE {
-                input_chunk.push(F::ZERO);
-            }
-
             // Add the inputs to our sponge state.
             for (i, &input) in input_chunk.iter().enumerate() {
                 self.sponge_state[i] = self.sponge_state[i] + input;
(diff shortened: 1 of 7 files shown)


src/plonk_challenger.rs
```diff
@@ -1,24 +1,39 @@
...
 }
 
+/// Observes prover messages, and generates verifier challenges based on the transcript.
+///
+/// The implementation is roughly based on a duplex sponge with a Rescue permutation. Note that in
+/// each round, our sponge can absorb an arbitrary number of prover messages and generate an
+/// arbitrary number of verifier challenges. This might appear to diverge from the duplex sponge
+/// design, but it can be viewed as a duplex sponge whose inputs are sometimes zero (when we perform
+/// multiple squeezes) and whose outputs are sometimes ignored (when we perform multiple
+/// absorptions). Thus the security properties of a duplex sponge still apply to our design.
```
daira, line 21: This works fine provided that the number of absorptions and squeezes in each round is fixed. Otherwise you need explicit padding, which I don't think your implementation has.

dlubarov: Hm I'm not seeing why a non-uniform round structure would be an issue; could you elaborate on that? This should be equivalent to a duplex modulo input padding and extra (ignored) outputs. E.g. if a particular round has `r` inputs and `2r` outputs, we can view it as adding two duplexes, where the second duplex gets an input of zero.

daira: My point is just that the input to the hash has to be unambiguously parseable as a transcript of the interactive protocol, which might not be the case if you don't use padding and have a variable number of absorptions or squeezes in any round.

dlubarov: I think I see what you mean. If the prover can influence the length of the transcript, then our (implicit) padding function is not injective, since `pad(m) = pad(m || 0)` if `m`'s length is not a multiple of `r`. Is that the issue?

Currently the number of prover messages in a given round is fixed (for a fixed instance) for us, but I agree that we would need a different padding function like `pad10*` if we had a variable-length round.

dlubarov: Thanks for taking a look!

PR #54, opened 2020-07-10
Add the option to remove public input openings

wborgeaud:
Makes the `o_public_inputs` proof attribute optional. In the case where the public input openings are not provided, the prover subtracts the public input wires commitments to the proof's wire commitments, and the verifier adds them back.
This greatly reduces the proof size when there's many public inputs, at the expense of increasing the verifying time.


src/plonk_util.rs
```diff
@@ -312,10 +312,76 @@ pub fn halo_g<F: Field>(x: F, us: &[F]) -> F {
...
 }
 
+/// Computes the commitments (without randomness) `Comm(L_i)` for `i=0..num`, where `L_i` is the `i-th`
+/// Lagrange basis polynomial on the subgroup of order `order` of `C`'s scalar field,
+/// i.e., `L_i(g^j) = delta_ij`, with `g` a `order`-th primitive root of unity.
+pub fn precompute_lagrange_commitments<C: Curve>(
+    order_log: usize,
+    num: usize,
+    msm_precomputation: &MsmPrecomputation<C>,
+) -> Vec<ProjectivePoint<C>> {
```
wborgeaud, line 322: There's probably a better way to compute these. Here I just use the naive way of computing the Lagrange basis polynomial coefficients one by one.
Also this could be data that we can store in a file and use it as a trustless (since anyone can recompute it) SRS.

wborgeaud:
I changed the verification method to use a commitment to the public input quotient polynomial `(WiresPoly - PIPoly) / VanishingPI` instead.

src/plonk_recursion.rs
```diff
@@ -87,20 +91,20 @@ pub fn recursive_verification_circuit<
...
-        .collect();
+    let o_public_inputs: Option<Vec<Target>> = proof.o_public_inputs.as_ref().map(|pis| {
+        pis.iter()
+            .cloned()
+            .flat_map(|opening_set| opening_set.o_wires)
+            .collect()
+    });
 
-    verify_assumptions::<C, InnerC>(&mut builder, degree_pow, &public_inputs, &o_public_inputs);
+    // verify_assumptions::<C, InnerC>(&mut builder, degree_pow, &public_inputs, &o_public_inputs);
```
dlubarov, line 101: Was this failing?

wborgeaud: Yes I get an `index out of bounds` error in `verify_assumptions`. I uncommented it and will try to fix it.

PR #58, opened 2020-07-15
Fix most of the compiler warnings

dlubarov:


dlubarov:
Oh yeah, I think `ORDER_X2` should be used eventually (as it's used in the BLS field), and probably `pedersen_commit`? Not sure if the others will be used though. In any case, I'd be fine with deleting anything unused since it's easy to restore later :)

PR #62, opened 2020-07-23
Added plookup

wborgeaud:
Added Plookup support. Currently this can be used to prove that a set of committed values is a subset of a public lookup table. We will then be able to integrate this into Plonky and add these checks for some circuits where a lookup table can save lots of gates, e.g., bit operations.

### DIFF
--- a/plookup/.gitignore
+++ b/plookup/.gitignore
@@ -1,0 +1,2 @@
+/target
+Cargo.lock
--- a/plookup/Cargo.toml
+++ b/plookup/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "plookup"
+version = "0.1.0"
+authors = ["wborgeaud <williamborgeaud@gmail.com>"]
+edition = "2018"
+
+[dependencies]
+plonky = {path="../../plonky"}
+anyhow = "1.0.31"
+rand = "0.7.3"
+itertools = "0.9.0"
+num = "0.2.1"
--- a/plookup/README.md
+++ b/plookup/README.md
@@ -1,0 +1,4 @@
+# Plookup on Plonky
+
+### Notes
+Uses const generics and thus requires Rust nightly.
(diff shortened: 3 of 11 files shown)


plookup/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
```
dlubarov, line 1: Not needed now that this is in the plonky repo, I think

plookup/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "plookup"
+version = "0.1.0"
+authors = ["wborgeaud <williamborgeaud@gmail.com>"]
+edition = "2018"
+
+[dependencies]
+plonky = {path="../../plonky"}
```
dlubarov, line 8: Can this be `path = ".."`? I sometimes use copies with other names, like `plonky-master` for a copy that I keep on master.

plookup/.gitignore
wborgeaud: Yes you're right. Actually, I'll just remove this `.gitignore`, since `Cargo.toml` is also in the plonky repo.

plookup/Cargo.toml
wborgeaud: Yes that works too. Made the change.

plookup/README.md
```diff
@@ -0,0 +1,4 @@
+# Plookup on Plonky
+
+### Notes
+Uses const generics and thus requires Rust nightly.
```
dlubarov, line 4: Seems fine for now but let's try to avoid other nightly stuff if we can, so that if it ends up blocking some projects from using Plonky, we can switch back to stable without too much work.

wborgeaud: Yes I used const generics mostly out of curiosity and because I think they fit well for this use case. But it should be pretty easy to get the same behavior with `Vec`s if we need to run on stable.

PR #63, opened 2020-07-26
Circuit code for bigint and foreign field arithmetic

dlubarov:
Currently foreign field addition takes ~400 gates and multiplication takes ~600 gates, most of which is range checking via `Base4SumGate`. The reduction is the most expensive part, so batch operations (like adding many products of foreign field elements) should be much more efficient, since they can do a single reduction. There's definitely room for improvement if we want to optimize further; I've just been focused on simplicity. Of couse plookup could also be integrated for much cheaper range checking.

A few design questions that came up:

- How should we fit bounded types into the type hierarchy? Originally I had `Target`, `BoundedTarget`, `BigIntTarget`, and `BoundedBigIntTarget`. It was getting a bit cluttered with bounded and unbounded versions of various methods, so I merged the latter two (so `BigIntTarget` always has a bound; it just might be a trivial bound based on the number of limbs). Ideally I'd like to also merge `Target` and `BoundedTarget`, but I don't see a clean way to do that since `Target` is an enum.

- Should we try to keep `num` types like `BigUInt` out of the public API? Originally I was aiming to, since other projects might be using a different version of `num`, and mixing versions can lead to confusing errors. But if we make `BigIntTarget` public, then we essentially need `BigUInt` in the public API.

- Should `ForeignField` be parameterized with a `Field` type? It wouldn't technically be used, so we'd need to add it with `PhantomData`, but it might increase type safety somewhat, ensuring that different foreign fields can't be added or what not. (The same applies to `AffinePointTarget`.)

### DIFF
--- a/src/circuit_builder.rs
+++ b/src/circuit_builder.rs
@@ -3,8 +3,9 @@ use std::collections::{BTreeMap, HashMap};
 use crate::gates::*;
 use crate::plonk_util::{commit_polynomials, halo_n, polynomials_to_values_padded, sigma_polynomials, values_to_polynomials};
 use crate::util::{ceil_div_usize, log2_strict, transpose};
-use crate::{blake_hash_base_field_to_curve, blake_hash_usize_to_curve, fft_precompute, generate_rescue_constants, msm_precompute, AffinePoint, AffinePointTarget, Circuit, Curve, CurveMsmEndoResult, CurveMulEndoResult, CurveMulOp, Field, HaloCurve, PartialWitness, PublicInput, Target, TargetPartitions, VirtualTarget, Wire, WitnessGenerator, NUM_CONSTANTS, NUM_WIRES};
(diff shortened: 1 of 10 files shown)


dlubarov:
> * I also don't see a clean way to merge `Target` and `BoundedTarget`. Maybe we could make `BoundedTarget` the new default and also have a trivial bound if the bound isn't used. I also think it should be ok to keep it as it is and have both `Target` and `BoundedTarget`.

I think it would be ideal if there was only one in the public API, though the enum makes it a little tricky. Maybe we could hide the enum and expose constructor methods instead, something like this?

```
pub struct Target {
  value: UnboundedTarget,
  max: BigUint,
}

impl Target {
  pub wire(gate: usize, input: usize) -> Self { ... }
  virtual(index: usize) -> Self { ... }
  pub bounded(&self, max: &BigUint) -> Self { ... }
}

enum UnboundedTarget {
    VirtualTarget(VirtualTarget),
    Wire(Wire),
}
```

It'd affect a lot of code though so I'll leave it for now.

> * I don't have a problem with having `BigUInt` in the public API. We could always write helper functions to pass the `max` as a string or limb vector for people not wanting to use `BigUInt`. We could also use the [uint](https://crates.io/crates/uint) crate by Parity. It should be lighter than `num` (for some reason it's not on crates.io) and is already used by some other crypto projects.

Sounds good -- let's leave it for now and see if any issues come up.

> * I think it would be nice to add the foreign field as a `PhantomData`. It's looks a bit weird now having the `ForeignFieldTarget` type not be linked to any actual foreign field. Same for `AffinePointTarget`.

I agree. Will add that as a small followup PR.

PR #64, opened 2020-07-27
New polynomial API

wborgeaud:
This PR adds a `Polynomial` struct that wraps a vector of coefficients. This provides a type-level distinction between polynomials in coefficient vs evaluation form, and more generally between a polynomial and a vector of field elements. It also allows for a more object-oriented design for handling polynomial, e.g. `p.add(&q)` instead of `polynomial_addition(&p, &q)`.

### DIFF
--- a/src/plonk_util.rs
+++ b/src/plonk_util.rs
@@ -1,6 +1,6 @@
 use crate::partition::get_subgroup_shift;
 use crate::witness::Witness;
-use crate::{fft_with_precomputation_power_of_2, ifft_with_precomputation_power_of_2, msm_execute_parallel, AffinePoint, CircuitBuilder, Curve, FftPrecomputation, Field, HaloCurve, MsmPrecomputation, ProjectivePoint, Target, NUM_ROUTED_WIRES};
+use crate::{fft_with_precomputation_power_of_2, ifft_with_precomputation_power_of_2, msm_execute_parallel, AffinePoint, CircuitBuilder, Curve, FftPrecomputation, Field, HaloCurve, MsmPrecomputation, Polynomial, PolynomialCommitment, ProjectivePoint, Target, NUM_ROUTED_WIRES};
 use rayon::prelude::*;
 
 /// Evaluate the polynomial which vanishes on any multiplicative subgroup of a given order `n`.
@@ -152,11 +152,8 @@ pub(crate) fn powers_recursive<C: HaloCurve>(
 }
 
 /// Returns the evaluation of a list of polynomials at a point.
-pub(crate) fn eval_coeffs<F: Field>(coeffs: &[Vec<F>], powers: &[F]) -> Vec<F> {
-    coeffs
-        .iter()
-        .map(|c| F::inner_product(c, &powers))
-        .collect()
+pub(crate) fn eval_polys<F: Field>(polys: &[Polynomial<F>], powers: &[F]) -> Vec<F> {
+    polys.iter().map(|p| p.eval_from_power(powers)).collect()
 }
 
 /// Zero-pad a list of `n` polynomial coefficients to a length of `8n`, which is the degree at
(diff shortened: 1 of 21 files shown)


src/plonk_util.rs
```diff
@@ -171,33 +168,38 @@ pub(crate) fn pad_to_8n<F: Field>(coeffs: &[F]) -> Vec<F> {
...
-    coefficients_vec
-        .iter()
-        .map(|coeffs| fft_with_precomputation_power_of_2(coeffs, fft_precomputation))
+    polys_vec
+        .par_iter()
+        .map(|poly| poly.eval_domain(fft_precomputation))
         .collect()
 }
 
 pub(crate) fn coeffs_to_values_padded<F: Field>(
```
dlubarov, line 192: Maybe `polynomials_to_values_padded` now?

src/polynomial.rs
```diff
@@ -1,343 +1,471 @@
...
+    pub fn from_coeffs(coeffs: &[F]) -> Self {
+        Self(coeffs.to_vec())
+    }
 
-pub fn trim<F: Field>(a: &mut Polynomial<F>) {
-    let d = degree(a);
-    a.drain(d + 1..);
-}
+    /// Returns the coefficient vector.
+    pub fn coeffs(&self) -> Vec<F> {
```
dlubarov, line 47: Maybe best to return a reference and let the caller determine if it needs to be cloned or not. I think it'd be equally idiomatic since Rust getters often return refs, from what I've seen.

src/polynomial.rs
```diff
@@ -1,343 +1,471 @@
...
+    }
+
+    /// Removes the coefficients in the range `range`.
+    fn drain<R: RangeBounds<usize>>(&mut self, range: R) {
+        self.0.drain(range);
+    }
+
+    /// Evaluates the polynomial at a point `x`.
+    pub fn eval(&self, x: F) -> F {
+        self.iter().rev().fold(F::ZERO, |acc, &c| acc * x + c)
```
dlubarov, line 99: Nice 😎

src/polynomial.rs
```diff
@@ -1,343 +1,471 @@
...
+    }
+
+    /// Negates the polynomial's coefficients.
+    fn neg(&self) -> Self {
+        Self(self.iter().map(|&x| -x).collect())
+    }
+
+    /// Removes leading zero coefficients.
+    pub fn trim(&mut self) {
+        if !self.is_zero() {
```
dlubarov, line 148: Maybe nice to have a `degree_plus_1` method (in addition to `degree`), for cases like this where calling `degree` could fail

src/plonk_util.rs
```diff
@@ -171,33 +168,38 @@ pub(crate) fn pad_to_8n<F: Field>(coeffs: &[F]) -> Vec<F> {
...
     fft_precomputation: &FftPrecomputation<F>,
 ) -> Vec<Vec<F>> {
-    coefficients_vec
-        .iter()
-        .map(|coeffs| fft_with_precomputation_power_of_2(&pad_to_8n(coeffs), fft_precomputation))
+    polys_vec
+        .par_iter()
+        .map(|poly| {
+            let mut padded_poly = poly.clone();
+            padded_poly.pad(poly.len() * 8);
```
dlubarov, line 200: Can this use `padded`?

src/polynomial.rs
```diff
@@ -1,343 +1,471 @@
...
+
+    /// Polynomial multiplication.
+    pub fn mul(&self, b: &Self) -> Self {
+        if self.is_zero() || b.is_zero() {
+            return Self::zero(1);
+        }
+        let a_deg = self.degree();
+        let b_deg = b.degree();
+        let mut a_pad = self.clone();
+        a_pad.pad(a_deg + b_deg + 1);
```
dlubarov, line 186: Can these two use `padded`?

wborgeaud:
Thanks! Made the changes.

PR #65, opened 2020-07-29
Remove public inputs from proof

wborgeaud:
This PR removes the `output_pis` flag used currently in the prover so that proofs no longer contain the openings at the public input gates.
Note: I haven't modified `ProofTarget` yet.

### DIFF
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -10,7 +10,7 @@ use crate::halo::batch_opening_proof;
 use crate::partition::{get_subgroup_shift, TargetPartitions};
 use crate::plonk_challenger::Challenger;
 use crate::plonk_proof::{OldProof, Proof};
-use crate::plonk_util::{commit_polynomials, eval_l_1, eval_polys, eval_zero_poly, pad_to_8n, permutation_polynomial, polynomials_to_values_padded, powers, reduce_with_powers, values_to_polynomials};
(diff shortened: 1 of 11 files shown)


src/plonk.rs
```diff
@@ -367,25 +329,15 @@ impl<C: HaloCurve> Circuit<C> {
...
-                zeta_sf * self.subgroup_generator_n,
-                zeta_sf * self.subgroup_generator_n.exp_usize(GRID_WIDTH),
-            ],
-        ]
+        let opening_points = [vec![
+            zeta_sf,
+            zeta_sf * self.subgroup_generator_n,
+            zeta_sf * self.subgroup_generator_n.exp_usize(GRID_WIDTH),
+        ]]
         .concat();
```
dlubarov, line 341: Can just be a `vec!` now right?

src/plonk.rs
```diff
@@ -202,20 +195,11 @@ impl<C: HaloCurve> Circuit<C> {
...
-                    .map(|i| {
-                        (0..w.len())
-                            .map(|j| w[j][i] * alpha_sf.exp_usize(j))
-                            .fold(C::ScalarField::ZERO, |acc, x| acc + x)
-                    })
-                    .collect::<Vec<_>>(),
-            )
-        });
+        let vanishing_pis_poly =
+            scale_polynomials(wire_polynomials_no_pis, alpha_sf, self.degree());
```
dlubarov, line 199: Good idea 👍

src/plonk.rs
wborgeaud: Oh yes thanks! I changed it everywhere, but this one must have slipped through.

PR #67, opened 2020-08-04
Typed targets

wborgeaud:
Added field and curve types to target structs, e.g., `Target` is now `Target<F>` where `F: Field`.
This allows for compile-time checks that the code is handling values in the correct field.

### DIFF
--- a/src/circuit_curve.rs
+++ b/src/circuit_curve.rs
@@ -227,7 +227,6 @@ impl<C: HaloCurve> CircuitBuilder<C> {
         impl<C: HaloCurve, InnerC: HaloCurve<BaseField = C::ScalarField>>
             WitnessGenerator<InnerC::BaseField> for ResultGenerator<C, InnerC>
         {
-            // fn dependencies(&self) -> Vec<Target<C::ScalarField>> {
             fn dependencies(&self) -> Vec<Target<InnerC::BaseField>> {
                 vec![
                     self.mul.scalar.convert::<InnerC::BaseField>(),
--- a/src/circuit_curve.rs
+++ b/src/circuit_curve.rs
@@ -14,7 +14,9 @@ impl<C: Curve> AffinePointTarget<C> {
     }
 }
 
-/// Represents a scalar * point multiplication operation.
+/// Represents a scalar * point multiplication operation on `InnerC`.
+/// `scalar` is modelled here in the "wrong" field `InnerC::BaseField = C::ScalarField` for coherence.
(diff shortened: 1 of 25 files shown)


src/plonk_proof.rs
```diff
@@ -160,63 +160,65 @@ impl<C: HaloCurve> OldProof<C> {
...
     /// The purported value of G, i.e. <s, G>, in the context of Halo.
-    pub halo_g: AffinePointTarget,
+    pub halo_g: AffinePointTarget<InnerC>,
     /// The data used in the final Schnorr protocol of the Halo opening proof.
-    pub schnorr_proof: SchnorrProofTarget,
+    pub schnorr_proof: SchnorrProofTarget<InnerC>,
 }
 
-impl ProofTarget {
+impl<C: HaloCurve, InnerC: HaloCurve<BaseField = C::ScalarField, ScalarField = C::BaseField>>
```
dlubarov, line 212: Does it work OK without `ScalarField = C::BaseField`? I think it'd be best not to require that, in case someone wanted to do bounded recursion e.g. with a Cocks-Pinch curve.

src/circuit_curve.rs
```diff
@@ -210,21 +212,28 @@ impl<C: HaloCurve> CircuitBuilder<C> {
...
+            _phantom: PhantomData<C>,
         }
 
-        impl<InnerC: HaloCurve> WitnessGenerator<InnerC::BaseField> for ResultGenerator<InnerC> {
-            fn dependencies(&self) -> Vec<Target> {
-                vec![self.mul.scalar, self.mul.point.x, self.mul.point.y]
+        impl<C: HaloCurve, InnerC: HaloCurve<BaseField = C::ScalarField>>
+            WitnessGenerator<InnerC::BaseField> for ResultGenerator<C, InnerC>
+        {
+            // fn dependencies(&self) -> Vec<Target<C::ScalarField>> {
```
dlubarov, line 230: Can `C` be removed and this deleted?

src/plonk_proof.rs
dlubarov: (Also comes up a couple times below in `plonk_recursion.rs`)

src/circuit_curve.rs
dlubarov, line 230: Can be removed

src/plonk_proof.rs
wborgeaud: Good point, I made the changes.

PR #78, opened 2020-09-08
Expand test suite for field arithmetic

### LINKED ISSUE #76: Test suite for finite field arithmetic
Before embarking on refactoring (#68), we need a more thorough test suite for finite field arithmetic. Many of the existing tests only test internal consistency, and that only wrt randomly generated inputs. A more thorough test suite will read in a fixed set of independently generated inputs and outputs (so the tests are repeatable) that have been more deliberately chosen (so we know we hit boundary cases).

unzvfu:
Basic test suite for field arithmetic. Closes #76.


src/field/field.rs
```diff
@@ -486,3 +486,212 @@ macro_rules! test_square_root {
...
+
+    /// Apply the unary functions `op` and `expected_op`
+    /// coordinate-wise to the inputs from `test_inputs(modulus,
+    /// word_bits)` and panic if the two resulting vectors differ.
+    pub fn run_unaryop_test_cases<T, UnaryOp, ExpectedOp>(
+        modulus: &BigUint, word_bits: usize,
+        op: UnaryOp, expected_op: ExpectedOp)
+        -> Result<()>
+    where
+        T: Field,
```
dlubarov, line 550: nit: we usually use `F` for Field type params (ditto for `run_binaryop_test_cases`)

src/field/field.rs
```diff
@@ -486,3 +486,212 @@ macro_rules! test_square_root {
...
+    ($field:ty) => {
+        use crate::field_tests;
+        use crate::field_to_biguint;
+        use std::io::Result;
+        use std::ops::{Add,Sub,Mul,Neg,Div};
+        use num::{BigUint, Zero, One};
+
+
+        /// Return the modulus of the type `Fld`.
+        fn field_modulus<Fld>() -> BigUint
```
dlubarov, line 610: nit: we usually use `F` for Fields. Also could compact it a bit with `F: Field` instead of the where clause.

src/field/field.rs
```diff
@@ -486,3 +486,212 @@ macro_rules! test_square_root {
...
+        where Fld: Field {
+            field_to_biguint(Fld::T) * (BigUint::one() << Fld::TWO_ADICITY) + 1u32
```
dlubarov, line 612: Could also do `field_to_biguint(F::NEG_ONE) + 1u32` as a possible alternative, but this is fine too if you prefer it.

src/field/field.rs
```diff
@@ -486,3 +486,212 @@ macro_rules! test_square_root {
...
+        }
+
+        // Can be 32 or 64; doesn't have to be computer's actual word
+        // bits. Choosing 32 gives more tests...
+        const WORD_BITS: usize = 32;
+
+        // TODO: I've given these unreasonably long names to avoid
+        // name conflicts (notably with the 'negation()' test for
+        // Bls12377Base. It would be better that these tests were
```
dlubarov, line 621: That's a good idea. I think it should work fine like

```
macro_rules! test_arithmetic {
    ($field:ty) => {
        mod arithmetic {
            ...
```

and besides method name collisions, it might help avoid redundant imports (not sure if those would cause compiler warnings or what).

src/field/field.rs
```diff
@@ -486,3 +486,212 @@ macro_rules! test_square_root {
...
+                |x, y| x * y % &modulus)
+        }
+
+        #[test]
+        fn arithmetic_square() -> Result<()> {
+            let modulus = field_modulus::<$field>();
+            field_tests::run_unaryop_test_cases(
+                &modulus, WORD_BITS,
+                |x| <$field>::mul(x, x),
+                |x| x.clone() * x % &modulus)
```
dlubarov, line 668: nit: could be `&x * &x`

src/field/field.rs
```diff
@@ -486,3 +486,212 @@ macro_rules! test_square_root {
...
+        }
+
+        #[test]
+        #[ignore]
+        fn arithmetic_division() -> Result<()> {
+            // This test takes ages to finish so is #[ignore]d by default.
```
dlubarov, line 674: https://github.com/rust-num/num-bigint/issues/60 should make this better some day :)

dlubarov (approved):
Looks good to me apart from those minor things 👍

src/field/field.rs
```diff
@@ -486,3 +486,212 @@ macro_rules! test_square_root {
...
+            .collect();
+        // Inputs 'difference from' modulus value
+        let diff_mod = basic_inputs.iter()
+            .filter(|x| *x < &modulus && !x.is_zero())
+            .map(|x| modulus - x)
+            .collect();
+        let basics = basic_inputs.into_iter().filter(|x| x < &modulus).collect::<Vec<BigUint>>();
+        [basics, diff_max, diff_mod].concat()
+
+        // // There should be a nicer way to express the code above; something
```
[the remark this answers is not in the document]
unzvfu: I tried a few different things, but couldn't get everything to line up. Probably not worth spending more time on at the moment.

src/field/field.rs
unzvfu: Shorter is better! :)

src/field/field.rs
unzvfu: Wow, that's taking ages! I've added a comment referencing the issue.

src/field/field.rs
unzvfu: I wrapped the test functions in `mod arithmetic { ... }` and tried fixing up the various `use` statements and references to `$field` but I couldn't get this to work. Any clues as to what might be tripping me up?

dlubarov: Got it working here :) #79 

I wasn't quite sure what to do with `test_square_root()`; for now I just wrote an import inside the method. Would it make sense to merge `test_square_root!` into `test_arithmetic!`?

unzvfu: Yes, I think that makes sense. Feel free to do that in #79, otherwise I'm happy to do it here.

PR #79, opened 2020-09-10
Separate module for arithmetic tests

dlubarov:
There were a couple things causing compiler errors --

- The macro calls were using non-qualified names, as in `test_arithmetic!(TweedledumBase)`, but `TweedledumBase` is no longer in scope in the context of the new submodule. I switched to qualified names, as in `test_arithmetic!(crate::TweedledumBase)`.
- I needed to add back the `Field` imports in modules like `bls12_377_base::test`. Previously it was pulled in by `test_arithmetic!`, but not anymore since the macro imports it into a submodule.


unzvfu (approved):
Looks good!

Issue #68, opened 2020-08-09
Refactor existing modular arithmetic code


unzvfu:
There is currently a fair amount of copy-pasta between the different base field implementations (e.g. [here](https://github.com/mir-protocol/plonky/blob/master/src/field/tweedledee_base.rs#L79), [here](https://github.com/mir-protocol/plonky/blob/master/src/field/tweedledum_base.rs#L79) and [here](https://github.com/mir-protocol/plonky/blob/master/src/field/bls12_377_base.rs#L59)). This should be factored out, so as to more clearly and conveniently accommodate the coming implementations and specialisations.

NB: Need to be careful to maintain the performance advantages that some of those implementations have from inlining and specialisation.

This issue depends on the completion of #76 and #77.

dlubarov:
Agreed! I was imagining that those implementations would eventually diverge as we added field-specific assumptions, but it hasn't happened so far.

> NB: Need to be careful to maintain the performance advantages that some of those implementations have from inlining and specialisation.

That's a good point also. One small optimization that comes to mind is that `ORDER[2]` is zero for the Tweedle* curves. I assume the compiler is able to eliminate a mul and an add there, though I haven't verified that.

unzvfu:
One thing I didn't realise until I started trying to do the refactoring is that const generics [are only in Rust nightly](https://github.com/rust-lang/rust/issues/44580#issuecomment-662543117), not beta or stable. AFAICT, const generics are necessary in order to write generic code that works with type `[T; N]` where `N` is a parameter.

What's your policy on using features from nightly? I would usually be hesitant to jump ahead like that, but maybe you already use nightly for something else?

wborgeaud:
Hey Hamish, I used const generics for Plookup, also to get types of the form `[T; N]`. I think these types would fall under the MVP discussed [here](https://internals.rust-lang.org/t/stabilizing-a-const-generics-mvp/12727), so it's likely that they would be stabilized before we have to use Plonky in production.
We can wait for Daniel's confirmation but on my end I think it's safe to use them.

unzvfu:
Thanks William, that's promising!

unzvfu:
> That's a good point also. One small optimization that comes to mind is that `ORDER[2]` is zero for the Tweedle* curves. I assume the compiler is able to eliminate a mul and an add there, though I haven't verified that.

[...]

PR #80, opened 2020-09-20
Issue 68: Refactor existing modular arithmetic code

unzvfu:
This PR addresses the easy parts of #68. Almost all of the functions that were specialised to 4 or 6 limbs have been replaced with a single version that takes a const generic parameter `N` that can be any positive integer. There are a lot of changes, but they are all very repetitive and straight forward I think.

I say "almost all" of the functions because it became clear that more significant restructuring will be necessary to refactor all the repeated `montgomery_multiplication` implementations. Specifically, the dependence on the representation is more explicit in that function (though it is of course there in the other arithmetic functions). I have a solution in mind for tidying this all up that I will implement eventually (it's already partly done), but at this stage I think it's a better idea for me to do some actual performance work on the arithmetic. :)

Speaking of performance, it's important that we don't introduce any performance regressions with these refactoring PRs. I couldn't detect any performance issues with this PR with my setup, but the reviewer should double-check this on their setup!

(The commit log incorporates all the commits from my previous PR, because this one was built on top of that; the messages can be ignored, it will merge fine. Sorry for the noise; I'll rebase properly next time.)


benches/bigint_arithmetic.rs
```diff
@@ -2,26 +2,22 @@ use criterion::{black_box, Criterion};
 use criterion::criterion_group;
 use criterion::criterion_main;
 
-use plonky::{cmp_6_6, mul_6_6};
+use plonky::{cmp, mul};
```
dlubarov, line 5: I'm seeing a couple errors here -- it looks like `cmp` is not visible in this context, and `mul` is missing? (A normal `cargo build` wouldn't detect these, but we can use `cargo bench --no-run` to build the benchmark code.)

Re visibility, I think the compiler sees benchmarks as a separate crate, so AFAIK the only easy solution is to make `cmp` `pub`, though it's not ideal. Hopefully the compiler will change that at some point.

dlubarov (commented):
Looks great otherwise! I'll just do a quick `cargo bench u64` once I'm able to build, to double check for perf changes as you suggested, though it seems like there shouldn't be any.

benches/bigint_arithmetic.rs
unzvfu: Good catch! Quite right, I was just `cargo test`ing so I missed these errors. (I removed `mul` because the only time it was used was in the benchmarking code.)

And you're right about the visibility: In the previous version, `cmp_4_4` was declared `pub (crate)` but `cmp_6_6` was declared `pub`; in this refactor I used `pub (crate)` which broke the access to `cmp` in the benchmark.

Given that we would prefer not to have `cmp` public, does it make sense to put the benchmark in the crate somehow? The benchmark currently only has `cmp` tests, so it's pretty sparse; presumably many functions that we will want to benchmark in the future shouldn't be `pub` either.

dlubarov: I agree having benchmarks inside the crate would be cleaner, but it seems it isn't currently possible with Criterion.rs, as per https://github.com/bheisler/criterion.rs/issues/158 --

> Cargo does not expose any way for Criterion.rs to define benchmarks in the main crate (ie. in the src/ directory).

It seems like our options are
- Switch from Criterion.rs to Rust's native benchmark framework, though it's nightly-only
- Move functions like `cmp` to a separate crate, so they can be `pub` without appearing in Plonky's docs
- Just make the functions `pub` for now until something changes, like native benchmarking becoming stable
- Make these functions public only if a feature is enabled, as suggested [here](https://www.reddit.com/r/rust/comments/9rnuhi/what_should_i_use_today_for_benchmarking_in_rust/e8ii3pk/?utm_source=reddit&utm_medium=web2x&context=3)

I don't really like the last option, since new contributors would have to consult the readme to know how to run benchmarks, vs just running `cargo bench`. The others all seem reasonable to me -- what do you think? (Also @wborgeaud -- let us know if you have a preference.)

unzvfu: I think the easiest way forward is to take option 3: make `cmp` (and perhaps similar functions) public for now; we can keep an eye on improvements to Criterion and Rust's native benchmark and fix the visibility when they catch up.

In fact, I suspect that most (all?) of the `bigint_arithmetic` functions are benchmarked just as well indirectly via the public interfaces of the various concrete field types. For example, there's little point in benchmarking `bigint_arithmetic::add_no_overflow` since it's covered by the various `Add` trait implementations for Tweedle{dee,dum}, etc. So one option is to remove the `bigint_arithmetic` benchmarks altogether. Thoughts?

dlubarov: Hm in theory, if we were optimizing a private helper method, it seems like a bench that only calls that method might give us the best signal-to-noise ratio. Though in practice, for tiny methods like `cmp`, maybe that bench data could be misleading? Since we wouldn't be seeing how changes to `cmp` affect the pipelines and register allocations of callers.

I agree option 3 sounds good so that we can quickly unblock your change, and then let's decide on the more general convention in our next call.

unzvfu: Ok, committed option 3 solution. Yep, let's discuss next call.

dlubarov (approved):
LGTM! I ran the u64 benches a few times on your branch & master, and saw some 5-10% swings in both directions. I don't think performance was actually affected; most likely my laptop is just being inconsistent.

unzvfu:
So after all that, I am now somewhat doubtful that rustc optimises const generic code properly. I finished implementing Montgomery squaring last night and saw that, with const generics the code was _almost twice as slow_ as Montgomery multiplication, whereas when I simply replaced the const generic parameter with the number 4 (for the Tweedle* curves) in the same code, it took only about 60% of the time of Montgomery multiplication which is much more in line with expectations (and a pretty great result itself!).

I'll have to investigate a bit further to see if other people have had performance problems with const generics; I assumed they would work much like non-type template parameters in C++, but there's no way we should be seeing such huge performance differences if that were the case. Our options might be (i) return to the previous situation with explicit specialisations to limb length 4 and 6, or (ii) use macros instead. Let's discuss further at the next catch-up.

dlubarov:
Interesting! Sorry I didn't think of this in advance, but I think the issue may be with `#[unroll_for_loops]`. (It's from this [crate](https://crates.io/crates/unroll), not something officially supported by Rust devs. There's a [ticket](https://github.com/rust-lang/rfcs/issues/2219) for official support, but it probably won't happen soon.) I think it only looks for for loops with integer literal bounds.

Maybe we could build a proc macro similar to `#[unroll_for_loops]`, which replaced a for loop like

```rust
#[unroll_const_generics(max = 6)]
for i in 0..n {
  body(i);
}
```

with

```rust
if 0 < n { body(0); }
if 1 < n { body(1); }
if 2 < n { body(2); }
if 3 < n { body(3); }
if 4 < n { body(4); }
if 5 < n { body(5); }
if 6 < n { body(6); }
// Fallback in case a loop has more than 6 iterations.
for i in 6..n {
  body(i);
}
```

Then I think we should end up with the unrolled code we want, after the compiler eliminated the dead code and the always-true conditionals. Admittedly it would be a huge hack, and it might slow down compilation (though probably not that bad for small-ish `max` values like 6).

dlubarov:
We actually started with macros, but eventually got rid of them [here](https://github.com/mir-protocol/plonky/commit/ae70b77e6c590f4bcb0448c1d6fd14071ab3cae6). As I recall we ran into a couple little annoyances. One was that some of our macros would produce dead code given certain parameters, e.g. in subtraction, we had a sometimes-empty for loop to check limbs which only existed in one operand. (Of course we could have written a separate macro for subtraction with different limb counts, or suppressed the warnings.)

Another little thing was that we ended up with lists of macro invocations like
```
sub_symmetric!(sub_4_4, 4);
sub_symmetric!(sub_6_6, 6);
sub_asymmetric!(sub_7_6, 7, 6);
sub_asymmetric!(sub_5_4, 5, 4);
```
which were slightly annoying to maintain.

Nothing that major though, so we could definitely consider returning to that approach.

Perhaps another option to consider would be just removing the BLS curve so that we only need to deal with ~256 bit fields. There's a chance that we might go back to BLS (Aztec and Matter Labs are going that direction), but I think it's relatively low.

dlubarov:
I think this seems OK to merge. I was a bit confused since our main method (`cargo run --release`) seems to fail earlier on this branch than on master, but after digging into the error, it looks like it's just due to a recent fix on master (e30ffb85). (There are still other errors even on master though.)

unzvfu:
I investigated a bit more on this topic. It seems rustc does unroll loops with const generic parameters to some extent, but for some reason my squaring code does not pass the compiler's internal threshold for code that's suitable to unroll. I'm fairly confident that the simple functions like `cmp`, `add`, _etc._ will be unrolled correctly with const generics (so this PR can indeed be merged), but anything more complicated will have to be dealt with manually.

Indeed, the `unroll_for_loops` procedural macro doesn't touch loops with const generic bounds, so we would have to resolve this in a manner similar to what you [mentioned above](https://github.com/mir-protocol/plonky/pull/80#issuecomment-697809468). At this stage, I think it will be more advantageous to simply focus on specialising to 4 limbs; then we can adapt to 6 or some other number of limbs if and when the need arises. My sense is that this foray into writing const generic versions of the code has been more of a distraction than helpful unfortunately.

Issue #74, opened 2020-08-09
Implement and benchmark non-reduced representation of base field elements


unzvfu:
It can be advantageous to leave a few bits of space in the representation of an element of a base field and to allow a certain number of overflows into that space during calculations. This can reduce branching, carry-chain dependencies, and, most importantly, the number of modular reductions in arithmetic operations.

PR #81, opened 2020-09-23
Issue 70: Fast modular squaring

### LINKED ISSUE #70: Fast modular squaring
For sufficiently large numbers, squaring a number can be made up to ~40% faster than simply multiplying it with itself (for example, see [here](https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/large-integer-squaring-ia-paper.pdf)). We should investigate whether we can obtain some performance benefit for numbers of the size used in Plonky.

~Note that the existing Montgomery modmul implementation is CIOS (i.e. interleaves multiplication and reduction). To implement fast squaring we first need Montogomery REDC and we lose the (mild) performance benefits of CIOS layout.~
We can combine the fast squaring method with modular reduction with the CIOS layout in a fairly straight forward manner. This is mentioned (but not explained) in [Acar's thesis, Section 2.3](https://www.microsoft.com/en-us/research/wp-content/uploads/1998/06/97Acar.pdf). Another useful reference is [this post by the Goff developers](https://hackmd.io/@zkteam/modular_multiplication) (that also includes a cute trick to skip some additions for some moduli).

### LINKED ISSUE #86: Refactor the underlying representation of the two Tweedle curves
The two Tweedle curves have almost identical implementations. There should be a cleaner way to express that; I'm thinking something like a 'Montgomery256' trait that provides Montgomery representation and arithmetic for moduli up to 256 bits.

This issue follows on from #68.

unzvfu:
This PR implements faster modular squaring using the Montgomery CIOS method. The improvement over multiplication comes from observing that only half of the 'off-diagonal' pairs of digits need to be multiplied because of the symmetry of the input.

Main references were [Acar's PhD thesis](https://www.microsoft.com/en-us/research/wp-content/uploads/1998/06/97Acar.pdf) and a [blog post by the Goff team](https://hackmd.io/@zkteam/modular_multiplication) (though I haven't yet implemented the improvement they are talking about at the latter link).

Benchmarking on my laptop suggests this implementation costs 60-70% of the time of the existing Montgomery multiplication function. Specifically multiplication takes around 48ns and squaring around 33ns. (Both drop by a further 5-10% when the conditional subtraction is removed at the end, which will be implemented as part of #74.)

Closes #70. 

ETA: I took the opportunity to refactor the Monty arithmetic from the two Tweedle curves in this PR, so it closes #86 too.

### DIFF
--- a/src/field/field.rs
+++ b/src/field/field.rs
@@ -647,7 +647,7 @@ macro_rules! test_arithmetic {
                 let modulus = field_modulus::<$field>();
                 field_tests::run_unaryop_test_cases(
                     &modulus, WORD_BITS,
-                    |x| <$field>::mul(x, x),
+                    |x| <$field>::square(&x),
                     |x| &x * &x % &modulus)
             }
 
--- a/src/field/mod.rs
+++ b/src/field/mod.rs
@@ -3,6 +3,7 @@ pub use bls12_377_scalar::*;
 pub use field::*;
 pub use tweedledee_base::*;
 pub use tweedledum_base::*;
+pub use monty::*;
 
 mod bls12_377_base;
 mod bls12_377_scalar;
@@ -10,3 +11,4 @@ mod bls12_377_scalar;
 mod field;
 mod tweedledee_base;
 mod tweedledum_base;
+mod monty;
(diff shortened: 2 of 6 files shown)


dlubarov (approved):
Nice cleanup!

PR #82, opened 2020-09-25
Serialization + Logging

wborgeaud:
This PR adds serialization for the `Proof` and `VerificationKey` structs using Serde's `(De)Serialization` traits. 
It also adds basic logging using the [pretty-env-logger](https://github.com/seanmonstar/pretty-env-logger) crate.
Finally I've cleaned up all Clippy lints.

### DIFF
--- a/src/plonk_proof.rs
+++ b/src/plonk_proof.rs
@@ -1,10 +1,11 @@
 use anyhow::{anyhow, Result};
+use serde::{Deserialize, Serialize};
 
 use crate::plonk_challenger::Challenger;
 use crate::plonk_util::{halo_g, halo_n, halo_s};
 use crate::{AffinePoint, AffinePointTarget, Curve, Field, HaloCurve, PartialWitness, Target, SECURITY_BITS};
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Serialize, Deserialize)]
 pub struct SchnorrProof<C: HaloCurve> {
     pub r: AffinePoint<C>,
     pub z1: C::ScalarField,
@@ -17,7 +18,7 @@ pub struct SchnorrProofTarget<C: Curve> {
     pub z2: Target<C::ScalarField>,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
 pub struct Proof<C: HaloCurve> {
     /// A commitment to each wire polynomial.
     pub c_wires: Vec<AffinePoint<C>>,
@@ -29,11 +30,11 @@ pub struct Proof<C: HaloCurve> {
     pub c_pis_quotient: AffinePoint<C>,
 
     /// The opening of each polynomial at `zeta`.
-    pub o_local: OpeningSet<C::ScalarField>,
+    pub o_local: OpeningSet<C>,
     /// The opening of each polynomial at `g * zeta`.
-    pub o_right: OpeningSet<C::ScalarField>,
+    pub o_right: OpeningSet<C>,
     /// The opening of each polynomial at `g^65 * zeta`.
-    pub o_below: OpeningSet<C::ScalarField>,
+    pub o_below: OpeningSet<C>,
 
     /// L in the Halo reduction.
     pub halo_l: Vec<AffinePoint<C>>,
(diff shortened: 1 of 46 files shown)


src/plonk_proof.rs
```diff
@@ -273,26 +274,26 @@ impl<C: HaloCurve, InnerC: HaloCurve<BaseField = C::ScalarField>> ProofTarget<C,
 }
 
 /// The opening of each Plonk polynomial at a particular point.
-#[derive(Clone, Debug)]
-pub struct OpeningSet<F: Field> {
```
wborgeaud, line 277: I've spent a good amount of time trying to make Serde work with `OpeningSet<F: Field>` without success. The compiler always barfs on weird trait bounds errors. All of our use cases always were of the form `OpeningSet<C::ScalarField>`, so hopefully changing to `OpeningSet<C: Curve>` won't make much of a difference.

src/serialization.rs
```diff
@@ -1,7 +1,9 @@
...
-};
+use crate::{AffinePoint, Curve, Field};
+use serde::de::Error as DeError;
+use serde::de::Visitor;
+use serde::ser::Error as SerdeError;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt;
 use std::io::{Error, ErrorKind, Read, Result, Write};
 
 pub trait ToBytes {
```
dlubarov, line 9: Would it remove sense to remove these now that we use serde's `Serialize`/`Deserialize`?

A slight complication is that I don't think we could `impl<F: Field> Serialize for F` due to coherence rules, but I think it would be reasonable to make `Serialize`/`Deserialize` supertraits of `Field`.

src/serialization.rs
```diff
@@ -99,6 +123,12 @@ mod test {
                 x.write(&mut buf[..])?;
                 let y = <$field>::read(&buf[..])?;
                 assert_eq!(x, y);
+
+                // Serde (de)serialization
+                let ser = bincode::serialize(&x).unwrap();
```
dlubarov, line 128: nit: could these unwraps use `?` here (and in the method below)?

src/serialization.rs
wborgeaud: So `AffineCurve` internally uses `To/FromBytes` to implement `(De)Serialize`. `Field` should too and I'm going to fix this now. I think it makes sense to keep these traits to have a canonical way to read/write encoded field elements / curve points without relying on Serde. But if you think it'd be cleaner to remove them, I can do that.

dlubarov: Oh ok, sounds good for now. In the long term I think we might end up with different serialization methods for perf reasons, particularly for Montgomery vs canonical field impls. (I think Hamish is planning to switch from Montgomery to canonical for the Tweedle curves.)

dlubarov (approved):

src/verifier.rs
```diff
@@ -11,14 +12,16 @@ use crate::{blake_hash_usize_to_curve, fft_precompute, msm_execute_parallel, msm
...
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct VerificationKey<C: HaloCurve> {
     pub c_constants: Vec<AffinePoint<C>>,
     pub c_s_sigmas: Vec<AffinePoint<C>>,
     pub degree: usize,
     pub num_public_inputs: usize,
     pub security_bits: usize,
+    #[serde(skip)]
```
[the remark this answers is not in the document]
wborgeaud: Oh yes you're right, I was a bit lazy with those. I can replace `#[serde(skip)]` with `#[serde(skip_serializing_if = "Option::is_none")]` and then add a method to change these attributes to `None` if someone wants to save space when serializing a vk?
After a quick test, it seems like I would get into the same trouble as in https://github.com/mir-protocol/plonky/pull/82#discussion_r494839509 with `OpeningSet<F: Field>` for `FftPrecomputation<F: Field>`. But this can be fixed by defining a new type `FftPrecomputationCurve<C: Curve> = FftPrecomputation<C::ScalarField>`. Would that work for you?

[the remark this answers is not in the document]
dlubarov: `#[serde(skip_serializing_if = "Option::is_none")]` sounds like a neat solution!

Hm I'd prefer to keep `FftPrecomputation<F: Field>`, but I'll take a look at that error now and see if we can work around it.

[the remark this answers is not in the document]
dlubarov: Oh yeah that error seems tricky. Feel free to merge like this and I'll look into it more later.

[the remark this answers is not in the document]
wborgeaud: Yeah I spent quite some time trying to solve it for `OpeningSet` but without success. Maybe I'll try asking on Stackoverflow with a minimal failing example. 
I'll merge now and try to come back later with a fix.

PR #83, opened 2020-09-25
Prevent Serde from adding redundant bounds on `F`

dlubarov:
Since `Field` is already a subtrait of `Serialize` and `DeserializeOwned`, we don't need Serde to add its own bounds like `where F: Serialize`. The redundant bounds would normally be harmless, but they cause an error due to a compiler bug: https://github.com/rust-lang/rust/issues/41617


wborgeaud (approved):
Cool, thanks for fixing that! I'll do a quick PR next to add the `#[serde(skip_serializing_if = "Option::is_none")]` to the verification key.

PR #85, opened 2020-09-26
Custom Eq for Polynomial

dlubarov:
To treat missing entries and zeros the same way. Also added a couple tests.


wborgeaud (approved):
Looks good to me!

PR #87, opened 2020-09-30
Move public inputs to the end of circuits

wborgeaud:
I've implemented the functionalities to put the public inputs at the end of the circuits and not at the beginning as it is now.
I haven't modified `plonk_recursion.rs` except to make it compile.

### DIFF
--- a/src/circuit_builder.rs
+++ b/src/circuit_builder.rs
@@ -1,4 +1,5 @@
 #![allow(clippy::type_complexity)]
+
 use std::collections::{BTreeMap, HashMap};
 
 use crate::gates::*;
@@ -33,32 +34,14 @@ impl<C: HaloCurve> CircuitBuilder<C> {
         }
     }
 
-    pub fn stage_public_input(&mut self) -> PublicInput<C::ScalarField> {
+    pub fn add_public_input(&mut self) -> Target<C::ScalarField> {
         let index = self.public_input_index;
         self.public_input_index += 1;
-        PublicInput {
-            index,
-            _field: PhantomData,
-        }
+        Target::PublicInput(PublicInput::new(index))
     }
 
-    pub fn stage_public_inputs(&mut self, n: usize) -> Vec<PublicInput<C::ScalarField>> {
-        (0..n).map(|_i| self.stage_public_input()).collect()
-    }
-
-    /// Add `PublicInputGate`s which enable public inputs to be routed. Should be called after all
-    /// `stage_public_input[s]` calls, but before any gates are added.
-    pub fn route_public_inputs(&mut self) {
-        debug_assert_eq!(
-            self.num_gates(),
-            0,
-            "Must be called before any gates are added"
-        );
-        let num_pi_gates = ceil_div_usize(self.public_input_index, NUM_WIRES);
-        for i in 0..num_pi_gates {
-            self.add_gate_no_constants(PublicInputGate::new(i * 2));
-            self.add_gate_no_constants(BufferGate::new(i * 2 + 1));
-        }
+    pub fn add_public_inputs(&mut self, n: usize) -> Vec<Target<C::ScalarField>> {
+        (0..n).map(|_i| self.add_public_input()).collect()
     }
 
     pub fn add_virtual_target(&mut self) -> Target<C::ScalarField> {
@@ -71,7 +54,7 @@ impl<C: HaloCurve> CircuitBuilder<C> {
         (0..n).map(|_i| self.add_virtual_target()).collect()
     }
 
-    pub fn add_virtual_point_target<InnerC: Curve<BaseField = C::ScalarField>>(
+    pub fn add_virtual_point_target<InnerC: Curve<BaseField=C::ScalarField>>(
         &mut self,
     ) -> AffinePointTarget<InnerC> {
         let x = self.add_virtual_target();
@@ -79,7 +62,7 @@ impl<C: HaloCurve> CircuitBuilder<C> {
         AffinePointTarget { x, y }
     }
 
-    pub fn add_virtual_point_targets<InnerC: Curve<BaseField = C::ScalarField>>(
+    pub fn add_virtual_point_targets<InnerC: Curve<BaseField=C::ScalarField>>(
         &mut self,
         n: usize,
     ) -> Vec<AffinePointTarget<InnerC>> {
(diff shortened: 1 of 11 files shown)


src/circuit_builder.rs
```diff
@@ -1071,13 +1053,40 @@ impl<C: HaloCurve> CircuitBuilder<C> {
...
         // values to ensure nothing is learned from the out-of-H openings.
         for _i in 0..3 {
             self.add_blinding_gate();
         }
 
+        // Append the public inputs at the end of the circuit and update the copy constraints.
+        let num_gates_without_pis = self.num_gates();
+        self.append_public_input_gates();
+        self.route_public_inputs(num_gates_without_pis);
+
```
wborgeaud, line 1089: I've chosen to put the public input gates before the padding gates. It looks better to me, but has the disadvantage that we need to add `num_gates_without_pis` to the `Circuit` and `VerificationKey` structs to recover the position of the public input gates.
I could easily change it the have the public input gates after the padding gates, in which case `num_gates_without_pis` wouldn't be needed.

src/gates/public_input.rs
```diff
@@ -64,35 +64,36 @@ impl<C: HaloCurve> Gate<C> for PublicInputGate<C> {
...
+            if witness.contains_target(targets[i_wire]) {
+                let value = witness.get_target(targets[i_wire]);
+                result.set_wire(
+                    Wire {
+                        gate: self.index + 1,
+                        input: i_advice,
+                    },
+                    value,
+                );
+            }
```
wborgeaud, line 96: It turns out this was broken since `generate()` wasn't called unless all wires of the public input gate were set. The code worked because of how `PartialWitness::set_public_input` worked. Since I've removed this function, the bug here appeared. 
The fix is rather hacky, but I didn't see a better way to do this without modifying the `WitnessGenerator` trait.

src/target.rs
```diff
@@ -38,14 +39,25 @@ pub enum Target<F: Field> {
...
     }
+
+    pub fn index(&self) -> usize {
+        match self {
+            Target::PublicInput(pi) => pi.index,
+            Target::VirtualTarget(v) => v.index,
+            Target::Wire(w) => w.gate,
+            Target::_Field(_, _) => unreachable!()
+        }
+    }
```
wborgeaud, line 60: This is currently used only to get the the public input index in `plonk_recursion.rs`. I hesitated to make it panic when `Target` isn't a public input. What do you think?

src/gates/public_input.rs
dlubarov: Ah yeah I see the problem with the old code, but I think this approach wouldn't work in all cases. I.e. since the generator doesn't declare any deps, it may be run "too early" and no-op. (The code that drives generators only executes each generator once, so there wouldn't be another chance for the generator to run again later.)

It seems like (unless we change the generator model) there's just no way for this to work with a single generator for a whole `PublicInputGate`. Some possible workarounds:
- We could change the gate model so that a gate can register multiple generators, rather than a gate "being" a generator. This seems pretty natural to me, though it might make the gate code somewhat verbose. Relatedly, this would mean that gates don't necessarily need to store their indices in the witness, which might make circuit definitions slightly cleaner.
- We could have `set_target` handle public input targets in a special way, adding two entries to the witness, like `set_public_input` did. Might need some other changes to make sure everything goes through `set_target`; I'd need to double check.

What do you think?

src/witness.rs
```diff
@@ -167,16 +168,30 @@ impl<F: Field> PartialWitness<F> {
...
-    }
-
     pub fn extend(&mut self, other: PartialWitness<F>) {
         for (target, value) in other.wire_values {
             self.set_target(target, value);
         }
     }
+
+    #[allow(clippy::needless_collect)]
+    // Replace all `PublicInput`-type targets by their corresponding `Wire`-type targets
```
dlubarov, line 178: nits: Let's make it `pub(crate)` and change `//` to `///`

src/witness.rs
```diff
@@ -167,16 +168,30 @@ impl<F: Field> PartialWitness<F> {
...
+            if let Target::PublicInput(pi) = t {
+                Some((Target::Wire(pi.original_wire(offset)), *v))
+            } else {
+                None
+            }
+        }).collect::<Vec<_>>();
+
+        self.wire_values.retain(|t, _| !matches!(t, Target::PublicInput(_)));
+
+        for (t, v) in new_pis.into_iter() {
```
dlubarov, line 191: `self.wire_values.extend(new_pis);` should also work I think?

src/gates/public_input.rs
dlubarov: Or on second thought, does `PublicInputGate` still need a generator at all? I was thinking `PartialWitness` could store only the new `PublicInput` targets, and the two associated `Wire`s wouldn't appear at all until the `PartialWitness` was converted to a `Witness`.

wborgeaud: Oh yeah you're right. I didn't take into account the cases where some `Target::PublicInput(_)` haven't been set before calling `Circuit::generate_witness`. I've added a test that was failing with the previous version (before commit 3db3dae). 

I've added a fix that looks in the witness for targets in buffer gates following a public input gate. And if some are found, adds the corresponding non-routable wire in the public input gate to the partial witness. 
It's in `PartialWitnesss::copy_buffer_to_pi_gate` and is called after each call to `Circuit::generate_copies` in the witness generation.
I think it's easier to do it this way, and I couldn't find a way to make it work by modifying `set_target` without having to change a lot of existing code.

> We could change the gate model so that a gate can register multiple generators, rather than a gate "being" a generator. This seems pretty natural to me, though it might make the gate code somewhat verbose. Relatedly, this would mean that gates don't necessarily need to store their indices in the witness, which might make circuit definitions slightly cleaner.

I think this would be a good idea, but would require a relatively large refactor. Maybe we can keep it for when we rewrite Plonky.

dlubarov: Hm I think there may still be a couple troublesome cases, maybe we could talk through it tomorrow. Also it looks like `copy_buffer_to_pi_gate` would make generation quadratic-time, which could be OK as a temporary fix but I think we should aim for quasi-linear long-term. (Currently I think it's linear under certain assumptions.)

It seems tricky to get all the cases right since public inputs have three possible forms in `PartialWitness` (`Target::PublicInput`, original wire, routable wire). I think it may be easiest if we only have `Target::PublicInput` in `PartialWitness` (and hide `original_wire`/`routable_target` so that the user can't interact with the other forms), and introduce the other forms when converting `PartialWitness` to `Witness` and `TargetPartitions` to `WirePartitions`.

wborgeaud: Hmm I'm not sure that the complexity changes. `copy_buffer_to_pi(copy_results)` runs in time linear in `|copy_results|` which is also the same as `witness.extend(copy_results)`.
But I agree that this is maybe not the best design. We can talk about it more tomorrow.

dlubarov: Ah yeah I misread that; I thought `copy_buffer_to_pi` was iterating through the "main" `PartialWitness`.

dlubarov (approved):

PR #90, opened 2020-12-14
Curve: double improvement

MrNeither:
According to the fact that output of the double operation should be AffinePoint and _affine -> projective -> affine_ transformation is resource-intensive changes were done to fix it. Also, the changes below include little improvements without rewriting results.

Old implementation:
* 11 additions and 12 multiplications for addition operation at projective coordinates;
* 2 multiplications and 1 inverse element search for transition from projective to affine coordinates.

New implementation:
* 1 inverse element search, 7 additions and 4 multiplications.

Speed comparison of the old and new implementations is presented below:

| version        | time (us)       |
| old    | 8.1654  |
| new      | **7.4816** |

### DIFF
--- a/benches/bls12_g1.rs
+++ b/benches/bls12_g1.rs
@@ -32,6 +32,10 @@ fn criterion_benchmark(c: &mut Criterion) {
         black_box(p1_projective).double()
     }));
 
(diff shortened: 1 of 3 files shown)


dlubarov:
Looks good, thanks!

PR #91, opened 2020-12-14
Field: cyclic order improvement

MrNeither:
Here we added a little improvement in cyclic group computations.
Every finite cyclic group of order n is isomorphic to the additive group of Z/nZ, so first matching of elements will be at initial element. Thus, here we don't use HashSet which include all intermediate elements, we just wait for x^k=FIELD::ONE


dlubarov:
Good idea; it looks good to me so I'll go ahead and merge. Thanks for your help!

PR #92, opened 2020-12-14
Field::exp improvement

MrNeither:
We have slightly accelerated the exponentiation algorithm. The Rust compiler does not unwrap a 64-range cycle, so a fixed check has no benefit. That's why **IF** statement has been taken out of cycle to improve performance.

Time measurements:
| version        | time (us)      |
| old    | 9.7351 |
| new      | **9.1447** |

image
image


dlubarov:
Looks good to me 👍

PR #93, opened 2020-12-14
Field: k-th root improvement

MrNeither:
We have added an improvement to avoid regularly transformation field_to_biguint for each iteration inside of cycle by using field_to_biguint transformation once outside of cycle.

Below are time measurements:
| version        | time (us)      |
| old    | 17.390 |
| new      | **12.291** |

### DIFF
--- a/src/field/field.rs
+++ b/src/field/field.rs
@@ -6,11 +6,11 @@ use std::hash::Hash;
 use std::ops::{Add, Div, Mul, Neg, Sub};
 
 use anyhow::{Error, Result};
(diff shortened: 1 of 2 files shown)


dlubarov:
Thanks, it looks good to me, I just added some minor changes --

- Moved the test to `test_arithmetic!` so that all fields will get it
- Added note about the `n=0` case, which this now skips (which is fine since it's never a solution)
- Removed note about the old code (I prefer for the comments just focus on the current code)

PR #94, opened 2020-12-14
Field: num bits improvement

MrNeither:
It isn’t required to check each bit of number from left to right to calculate the number of bits. You just need to check bits from right to left until you meet zero limbs. The changes below include little improvements without rewriting result.

Below are time measurements:
| version        | time (ns)      |
| old    | 188.41 |
| new      | **87.340** |


dlubarov:
Good idea -- I just pushed some minor changes, but I like the approach. Thanks!

PR #95, opened 2020-12-14
Add check discriminant

MrNeither:
We have added an additional check to prevent attacks in case when the discriminant is equal to zero. You can read more  [this](https://www.dima.unige.it/~morafe/MaterialeCTC/p80-menezes.pdf) and [this](https://mathworld.wolfram.com/EllipticDiscriminant.html)

### DIFF
--- a/src/curve/curve.rs
+++ b/src/curve/curve.rs
@@ -54,6 +54,11 @@ pub trait Curve: 'static + Sync + Sized + Copy + Debug {
         }
         Ok(res)
     }
+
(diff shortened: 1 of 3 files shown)


dlubarov:
Hm in general I don't think we should concern ourselves with curve safety, since the (external) code used to find these curves (e.g. the [pasta search](https://github.com/zcash/pasta)) has its own safety checks. But since this check is very simple I suppose it doesn't hurt to include it. Thanks!

PR #100, opened 2021-02-14
Implement support for the Pasta curves

unzvfu:
This PR implements support for the [Pasta curve pair](https://electriccoin.co/blog/the-pasta-curves-for-halo-2-and-beyond/) which provide some benefits over the Tweedle* curve pair.

This PR does not attempt to implement any of the performance improvements mentioned in the aforementioned reference; the benefits such are likely to be fairly modest.

Most of the work was just converting the published constants into Montgomery form and pasting them into the right slots. This means that there is some additional duplication with the other field and curve implementations; this will be refactored in due course. The only exception to this was the choice of the `ZETA` and `ZETA_SCALAR` parameters which are not mandated by the Pasta curves, just their properties; so a choice was made and set in the implementation.

### DIFF
--- a/src/curve/mod.rs
+++ b/src/curve/mod.rs
@@ -6,6 +6,8 @@ pub use curve_multiplication::*;
 pub use curve_summations::*;
 pub use tweedledee_curve::*;
 pub use tweedledum_curve::*;
+pub use pallas_curve::*;
+pub use vesta_curve::*;
 
 mod bls12_377_curve;
 #[allow(clippy::module_inception)]
@@ -16,3 +18,5 @@ mod curve_multiplication;
 mod curve_summations;
 mod tweedledee_curve;
 mod tweedledum_curve;
+mod pallas_curve;
+mod vesta_curve;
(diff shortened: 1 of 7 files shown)


dlubarov (approved):
Thanks, looks good!

I had forgotten how much work it was to work out all the constants and stuff. It would be nice if we could compute most of them at compile time, but I think it's tricky in Rust since compile time computations are so limited, e.g. we can't allocate `BigUInt`s.

We should probably unify the tests at some point (a la `test_arithmetic`), but I can do that afterward.

unzvfu:
> Thanks, looks good!

Thanks. :)

> I had forgotten how much work it was to work out all the constants and stuff. It would be nice if we could compute most of them at compile time, but I think it's tricky in Rust since compile time computations are so limited, e.g. we can't allocate `BigUInt`s.

Yes, I gave this some thought since it proved quite error-prone to work out which values needed to be converted to Montgomery form and which didn't, and to do the Montgomery calculation "manually" for each value and copy it across. I think it's probably possible to make Montgomery multiplication a `const fn` and get most of the functionality we'd want, but I'm not sure it's worth it at this stage, since you can't even use `for` in a `const fn` (tail recursion only).

> We should probably unify the tests at some point (a la `test_arithmetic`), but I can do that afterward.

Agreed. I'm also happy to do the refactoring; I guess we'll see who gets there first. :)

PR #103, opened 2021-10-19
Compile on nightly

gilescope:
few drive by clippys also.


dlubarov (approved):
Thanks for the fixes! LGTM.

We're working on a "v2" of plonky, so we consider this repo sort of deprecated, but glad to know it still compiles now.

gilescope:
looking forward to it. Was glad to find a plonk in rust.

